PR #43, opened 2020-10-14
[RFC] Update vector store name.

xingmingjie:


kito-cheng (approved):
LGTM, thanks for fix this :)

PR #46, opened 2020-11-04
higher level interface for control vxrm and vxsat

HanKuanChen:


eopXD:
With the rounding mode intrinsics we have and vxsat intrinsics that we will add in the future, we should not probably expose interface of these CSR-s.

PR #51, opened 2020-12-01
Add note to vundefined

kito-cheng:
Clarify any operation with `vundefined` would result an unpredictable value.

### DIFF
--- a/rvv-intrinsic-rfc.md
+++ b/rvv-intrinsic-rfc.md
@@ -564,7 +564,7 @@ This section lists all utility functions to help users program in V intrinsics e
 
 ### Vector Initialization<a name="vector-init"></a>
 
-These utility functions are used to initialize vector values. They could be used in masking intrinsics with *zero in output* and *don't care in output* semantics.
+These utility functions are used to initialize vector values. They could be used in masking intrinsics with *don't care in output* semantics.
 
 ```
 Example:
--- a/rvv-intrinsic-rfc.md
+++ b/rvv-intrinsic-rfc.md
@@ -575,8 +575,8 @@ vint8m1_t vundefined_i8m1()
 Note: Any operation with `vundefined_* ()` are undefined and unpredictable,
 the only recommended usage is used as maskedoff operand; an additional note is
 any operation with `vundefined_* ()` will got unpredictable result, e.g.
-`vxor(vundefined (), vundefined ())` will not guarantee result vector with
-zeros.
+`vxor(vundefined (), vundefined ())` and `vec a = vundefined(); vec b = vxor(a, a);`
+both are not guarantee result vector with zeros.
 
 ### Reinterpret between floating point and integer types<a name="reinterpret-float"></a>
 
--- a/rvv-intrinsic-rfc.md
+++ b/rvv-intrinsic-rfc.md
@@ -416,8 +416,6 @@ vint8m1_t vcompress_vm_i8m1(vbool8_t vs1, vint8m1_t maskedoff, vint8m1_t vs2)
 
 There are two additional masking semantics: *zero in output* semantic and *don't care in output* semantic. Users could leverage *merge in output* intrinsics to simulate these two additional masking semantics.
 
-`vundefined` will operate on `vl` elements.
-
 ```
 Example:
 
--- a/rvv-intrinsic-rfc.md
+++ b/rvv-intrinsic-rfc.md
@@ -572,6 +572,12 @@ Example:
 vint8m1_t vundefined_i8m1()
 ```
 
+Note: Any operation with `vundefined_* ()` are undefined and unpredictable,
+the only recommended usage is used as maskedoff operand; an additional note is
+any operation with `vundefined_* ()` will got unpredictable result, e.g.
+`vxor(vundefined (), vundefined ())` will not guarantee result vector with
+zeros.
+
 ### Reinterpret between floating point and integer types<a name="reinterpret-float"></a>
 
 These utility functions help users to convert types between floating point and integer types. The reinterpreter intrinsics only change the types of underlying contents. It is a nop operation.


zakk0610:
LGTM

rvv-intrinsic-rfc.md
```diff
@@ -602,14 +600,19 @@ This section lists all utility functions to help users program in V intrinsics e
...
 
 ```
 Example:
 
 vint8m1_t vundefined_i8m1()
 ```
 
+Note: Any operation with `vundefined_* ()` are undefined and unpredictable,
+the only recommended usage is used as maskedoff operand. e.g.
+`vxor(vundefined (), vundefined ())` will not guarantee result vector with
```
topperc, line 613: My example case was:

vec a = vundefined();
vxor(a, a);

So vundefined() was only called once but constant folding inside the compiler propagated undef to both inputs. The suggestion to use (freeze undef) still wouldn't have made vxor(vundefined (), vundefined ()) produce 0 as each input would be a unique freeze and therefore not equal.

HanKuanChen: How about
```
vec a;
vxor(a, a);
```
If the above result is 0, I think users will treat the following code also get 0.
```
vec a = vundefined();
vxor(a, a);
```

topperc:
This is undefined behavior because the C standard says accessing unitialized memory is undefined behavior.

```
vec a;
vxor(a, a);
```

kito-cheng:
I am not sure does it too strong or aggressive if we have a statement here to describe such operation will got unpredictable result, that will give the compiler freedom to do both way, either will be acceptable, 0 or random garbage value.

e.g. add one more sentence: any variable initialized with `vundefined ()` do not guarantee any algebra properties on that, for example `vec a = vundefined(); vec b = vxor(a, a);` b will result an unpredictable value instead of vector with zeros.

kito-cheng:
@rofirrim @ebahapo @jim-wilson I guess we need more input on this issue.

The background of this issue: 

In algebra level `a ^ a` and `a - a` should always result `0`, however what if `a` is come from `vundefined ()`? should we(or could we) treat that as undefined behavior like unitialized variable?
```
vec a = vundefined();
vec b = vxor(a, a);
```

My preference is let define that undefined on the document, and give compiler the freedom to optimizing out that like unitialized variable.

rvv-intrinsic-rfc.md
rofirrim: I wouldn't expect  

```
vec a = vundefined();
vxor(a, a);
```

be equivalent to

```
vxor(vundefined(), vundefined());
```

because I understand `vundefined()` as `vrandombits()` as a non-pure function (and no defined semantics is given to the bits returned, they can be the same, or different every time, etc). Under these semantics `freeze undef` in LLVM looks to me it should work, shouldn't it?

topperc: InstCombine folds (freeze undef) to 0 so we lose track of the fact that it was undefined. This results in an instruction to create a 0 vector. Which in a lot of places is probably good to prevent false dependencies on an out of order core. But for the purpose of using it to indicate an undefined input for a masked operation it doens't work.

rofirrim:
> In algebra level `a ^ a` and `a - a` should always result `0`, however what if `a` is come from `vundefined ()`? should we(or could we) treat that as undefined behavior like unitialized variable?

We should treat it as "random set of bits every time is invoked" which I think it maps with your understanding of "uninitialized variable".

I don't think it is reasonable to make this case undefined behaviour (otherwise `vundefined()` seems of little use).

rvv-intrinsic-rfc.md
kito-cheng: @Hsiangkai has try to implement with `freeze undef` but it seems result worse code gen when used in maskedoff operand...

Testcase:
```
#include <riscv_vector.h>
void foo(int32_t *a, int32_t *b, int len) {
  vbool32_t mask = vmset_m_b32();
  vint32m1_t va = vle32_v_i32m1_m_vl (mask, vundefined_i32m1 (), a, len);
  vse32_v_i32m1_m_vl (mask, b, va, len);
}
```
Without `freeze undef`:
```
foo:                                    # @foo
# %bb.0:                                # %entry
        csrr    a3, vl
        vsetvli a3, a3, e8,mf4,tu,mu
        vmset.m v0
        vsetvli a2, a2, e32,m1,tu,mu
        vle32.v v1, (a0), v0.t
        vse32.v v1, (a1), v0.t
        ret
```
With `freeze undef`:
```
foo:                                    # @foo
# %bb.0:                                # %entry
        csrr    a3, vl
        vsetvli a3, a3, e8,mf4,tu,mu
        vmset.m v0
        vsetvli a3, zero, e32,m1,tu,mu
        vmv.v.i v1, 0                                        # <----------------------------------- Extra 0 init instruction
        vsetvli a2, a2, e32,m1,tu,mu
        vle32.v v1, (a0), v0.t
        vse32.v v1, (a1), v0.t
        ret
```

kito-cheng:
> I don't think it is reasonable to make this case undefined behaviour (otherwise vundefined() seems of little use).

I think the only usage is using as a maskedoff operand?

rvv-intrinsic-rfc.md
rofirrim: Oh I see, this is disappointing. I wasn't expecting that fold.

ISTR for AVX-512 clang already emits a zero vector for the equivalent undefined intrinsic. I understand doing that would incur on the same issue of creating a vector of zeroes.

LLVM-specifically: perhaps we can have a `@llvm.rvv.vundefined.<ty>()` that we select as `IMPLICIT_DEF` in SelectionDAG? Now I'm wondering if we want it to have side-effects or not as SelectionDAG will CSE it if it ends being mapped to a node without chain (and that would introduce some false dependences that @topperc mentioned). I wonder if we can give it side-effects but then (re-)chain it to the entry token of the BB. Could that work @topperc?

topperc: I think if its chained to the entry node, then it would then become subject to CSE again. SelectionDAG doesn't exclude chain nodes from CSE. It only excludes a couple specific opcodes and nodes with with Glue.

We could keep it without side effects and replace it with undef if its used in a masked intrinsic? That would allow those to select IMPLICIT_DEF. And other usages can just select to 0? We also need to select the vtype to "mask agnostic" right? Otherwise we have a false dependency.

kito-cheng:
@rofirrim @topperc , are you OK with current change? current text is kind of leaving some gray zone on following case:

```
vec a = vundefined();
vec b = vxor(a, a);
```

Although I hope we can relax that on the spec level, so that compiler having some degree to choose how they implement (kind of implementation-defined in C Lang spec), but I am fine to leave that as gray zone at this PR, and create another PR to further  discussion on the above case.

topperc:
> @rofirrim @topperc , are you OK with current change? current text is kind of leaving some gray zone on following case:
> 
> ```
> vec a = vundefined();
> vec b = vxor(a, a);
> ```
> 
> Although I hope we can relax that on the spec level, so that compiler having some degree to choose how they implement (kind of implementation-defined in C Lang spec), but I am fine to leave that as gray zone at this PR, and create another PR to further discussion on the above case.

I would like to see the example "vxor(vundefined (), vundefined ())" replaced with "vec a = vundefined(); vec b = vxor(a, a);" The important thing to show is that there is unexpected behavior if you call vundefined() once and use the returned value in two places.

kito-cheng:
@topperc Updated, I keep the existing example and add `vec a = vundefined(); vec b = vxor(a, a);`, let me know if any other suggestion or concern about the description :)

topperc (commented):
LGTM

PR #52, opened 2020-12-07
Add section for switching vtype and keep same VL in a Loop

zakk0610:


kito-cheng:
I would suggest don't use `loop` in the description since we also must guarantee that even outside loop, but honestly I don't have idea which term would be appropriate for that? code sequence? control flow?

```
vsetvle16m4(n);
vfloat16m4_t vx = vle32_v_f16m4(ptr_x); 
vfloat32m8_t vy = vle32_v_f32m8(ptr_y); 
vfwmacc_vf_f32m8(vy, 2.0, vx);
vse32_v_f32m8(ptr_y, vy);
```

kito-cheng:
Maybe we also need an example for explicitly VL one?
```
// Not in loop but we still need to guarantee the vtype is correct.
vl = vsetvl_e16m4 (n);
vfloat16m4_t vx = vle32_v_f16m4_vl (ptr_x, vl); 
// No need another vsetvl_e32m8 (n), use vl come from e16m4 is fine, and compiler will guarantee the correctness of vtype setting.
vfloat32m8_t vy = vle32_v_f32m8_vl (ptr_y, vl); 
vfwmacc_vf_f32m8_vl(vy, 2.0, vx, vl);
vse32_v_f32m8_vl(ptr_y, vy, vl);
```

HanKuanChen:
> I would suggest don't use loop in the description since we also must guarantee that even outside loop

Agree. This is not a loop related topic. In addition, the example should be
```
vl = vsetvl_e16m4(n);
vfloat16m4_t vx = vle16_v_f16m4(ptr_x);
// vsetvl_e32m8(vl);	// keep the same vl but change vtype
vfloat32m8_t vy = vle32_v_f32m8(ptr_y);
// vsetvl_e16m4(vl);	// keep the same vl but change vtype
vfwmacc_vf_f32m8(vy, 2.0, vx);
// vsetvl_e32m8(vl);	// keep the same vl but change vtype
vse32_v_f32m8(ptr_y, vy);
```

rvv-intrinsic-rfc.md
```diff
@@ -726,3 +727,32 @@ vuint8m1_t vreinterpret_u8 (vint8m1_t src);
...
+// vsetvl_e16m4(vl); // keep the same vl without change vtype manually
+vfwmacc_vf_f32m8(vy, 2.0, vx);
+// vsetvl_e32m8(vl); // keep the same vl without change vtype manually
+vse32_v_f32m8(ptr_y, vy);
+```
+
+```
+vl = vsetvl_e16m4(n);
+vfloat16m4_t vx = vle16_v_f16m4(ptr_x, vl);
+// vsetvl_e32m8(vl); // change vtype manually
```
kito-cheng, line 749: > change vtype manually

Maybe `No need to change vtype manually` ?

rvv-intrinsic-rfc.md
```diff
@@ -726,3 +727,32 @@ vuint8m1_t vreinterpret_u8 (vint8m1_t src);
...
+vfloat32m8_t vy = vle32_v_f32m8(ptr_y, vl);
+// vsetvl_e16m4(vl); // change vtype manually
+vfwmacc_vf_f32m8(vy, 2.0, vx, vl);
+// vsetvl_e32m8(vl); // change vtype manually
+vse32_v_f32m8(ptr_y, vy, vl);
+
+This example has a `vl` computed from `vsetvle16m4_t`, and changing the type to `vfloat32m8_t` in the middle.
```
kito-cheng, line 756: `vsetvle16m4_t` -> `vsetvl_e16m4`

rvv-intrinsic-rfc.md
```diff
@@ -726,3 +727,32 @@ vuint8m1_t vreinterpret_u8 (vint8m1_t src);
...
+With compiler's helping, users don't need to change vtype manually because `vfloat16m4_t` and `vfloat32m8_t` have the exact same number of elements (same `SEW`/`LMUL` ratio).
+Noted that when changing vtype with different `SEW`/`LMUL` ratio in a loop, the result is undefined.
```
kito-cheng, line 758: It seems will result an illegal-instruction exception more than undefined.
https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc#61-vsetvlivsetvl-instructions
```
...
Use of the instruction with a new SEW/LMUL ratio that would result in a
change of VLMAX is reserved. Implementations may set vill in this case.
...
```

Once `vill` is raised:
https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc#334-vector-type-illegal-vill
```
If the vill bit is set, then any attempt to execute a vector instruction that depends
upon vtype will raise an illegal-instruction exception.
```

> Noted that when changing vtype with different `SEW`/`LMUL` ratio in a loop, the result is undefined.

I worry about this wording is too strong, for example, with the reduction operation, we might change the vtype with VLMAX change within a loop, and according this statement the result is undefined since we've changing `SEW`/`LMUL` ratio within a loop.

Ref:
https://github.com/sifive/sifive-dsp-internal/blob/master/DSP/DSP-rvv/Source/FilteringFunctions/riscv_lms_f32.c#L122

zakk0610: I changed the description, hope it will be more clear.

kito-cheng (approved):
Thanks, that's LGTM.

PR #57, opened 2021-01-06
Add LMUL truncate and extension functions

zakk0610:


rvv-intrinsic-rfc.md
```diff
@@ -609,6 +610,25 @@ Example:
 vint64m1_t vreinterpret_v_i32m1_i64m1(vint32m1_t src)
 ```
 
+### LMUL truncation and LMUL extension functions<a name="lmul-trunc)"></a>
+
+These utility functions help users to truncate or extent current LMUL under same SEW regardless of vl (keep current vl).
```
kito-cheng, line 615: Do you mind having more description on `LMUL extension`? Maybe `The value for extension part of variable will be undefined value.`?

> same SEW regardless of vl (keep current vl).
I assume that means `vl` register won't affected/changed? if so, maybe ` (keep current vl).` ->  `, and won't change content of vl register `?

jan-wassenberg:
@kito-cheng Can we generalize this a bit more? For example: v8..v15 are an m8, then we can use trunc to get v8..v11 as an m4, but how do we access the upper half of m8 (v12..v15) - maybe upper(m8)->m4?

Similarly, can we concatenate two m4 into an m8?

kito-cheng:
> @kito-cheng Can we generalize this a bit more? For example: v8..v15 are an m8, then we can use trunc to get v8..v11 as an m4, but how do we access the upper half of m8 (v12..v15) - maybe upper(m8)->m4?

I think this can be done with vslidedown and LMUL trunc to get the upper half:
```
vint64m2_t vhighpart_i64m1_i64m1 (vint64m1_2 op1)
{
   size_t vl = vsetvlmax_e64m2 ()
   vint64m2_t tmp = vslideup_vx_i64m2(new_op1, new_op2, vl);
   return vlmul_trunc_v_i64m2_i64m1 (tmp);
}
```
But I guess there is room to improve on the compiler side for optimize out the slideup.

> Similarly, can we concatenate two m4 into an m8?

This can be done by vslideup, @zakk0610 has provide an implementation on that  https://github.com/riscv/rvv-intrinsic-doc/issues/54#issuecomment-751937378

jan-wassenberg:
@kito-cheng Thanks, makes sense. (Typo in the code: slideup->slidedown)

Looks like ext/trunc is all we need :)

kito-cheng (approved):
LGTM

PR #72, opened 2021-03-18
Use size_t for shift amount

zakk0610:
address https://github.com/riscv/rvv-intrinsic-doc/issues/68
we use size_t type temporarily before we have uint_xlen_t type.


kito-cheng:
LGTM

Issue #81, opened 2021-04-21
Note on Unit-stride Fault-Only-First-Loads is incorrect


topperc:
The note for these instructions says

```
Execute a regular load and stop load operation if there is 0 value of element, and then set vl equals the index of last non-zero value.
```

This is incorrect, the instruction does not look at the data returned. The instruction traps if the first element to be loaded is at an invalid address, for example it is a null pointer or it points to a page you don't have read access to. If the address of the first element is valid, but some other element being loaded is not valid, VL will be changed to the number of elements that were able to be loaded without a trap. It is up to the user to look for 0s or another terminator value(s) in the data returned up to the new vl value. The instruction is just allowing elements to be speculatively read many elements without knowing whether they will contain a terminator or not.

topperc:
Fixed by #82

PR #82, opened 2021-04-22
Update the note of unit-stride fault-only-first loads.

Hsiangkai:
Address issue #81.

### DIFF
--- a/rvv-intrinsic-api.md
+++ b/rvv-intrinsic-api.md
@@ -96,7 +96,7 @@ void vwrite_csr(enum RVV_CSR csr, unsigned long value);
 #### [Intrinsic functions list](intrinsic_funcs/02_vector_loads_and_stores_functions.md#unit-stride-fault-only-first-loads-functions)
 
 #### Notes
--  The instruction traps if the first element to be loaded is at an invalid address, for example it is a null pointer or it points to a page you don't have read access to. If the address of the first element is valid, but some other element being loaded is not valid, VL will be changed to the number of elements that were able to be loaded without a trap. It is up to the user to look for 0s or another terminator value(s) in the data returned up to the new vl value.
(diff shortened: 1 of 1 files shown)


zakk0610 (approved):
LGTM

rvv-intrinsic-api.md
```diff
@@ -96,7 +96,7 @@ void vwrite_csr(enum RVV_CSR csr, unsigned long value);
 #### [Intrinsic functions list](intrinsic_funcs/02_vector_loads_and_stores_functions.md#unit-stride-fault-only-first-loads-functions)
 
 #### Notes
-- Execute a regular load and stop load operation if there is 0 value of element, and then set vl equals the index of last non-zero value.
+-  The instruction traps if the first element to be loaded is at an invalid address, for example it is a null pointer or it points to a page you don't have read access to. If the address of the first element is valid, but some other element being loaded is not valid, VL will be changed to the number of elements that were able to be loaded without a trap. It is up to the user to look for 0s or another terminator value(s) in the data returned up to the new vl value.
```
topperc, line 99: I'm not sure if that's a complete list of reasons a load can trap.

Hsiangkai: Should we list all of the reasons of exceptions? Or could we just describe the behavior under the exceptions?

topperc: We can probably just use this from the V extension

```
The unit-stride fault-only-first load instruction is used to vectorize loops with data-dependent exit conditions (while loops). These instructions execute as a regular load except that they will only take a trap on element 0. If an element > 0 raises an exception, that element and all following elements in the destination vector regis- ter are not modied, and the vector length vl is reduced to the number of elements processed without a trap.
```

topperc (approved):
LGTM

PR #87, opened 2021-05-21
Add notes that vslideup/vslidedown and vcompress will use tail undist…

topperc:
…urbed policy.

These intrinsics are defined with a destination operand so they
should use tail undisturbed.

Other intrinsics are more ambiguous about what they use for tail
policy.


zakk0610:
LGTM

zakk0610 (approved):

PR #89, opened 2021-05-21
Update saxpy and sgemm examples to move the vsetvli into the loop bod…

topperc:
…y and not make the loop condition depend on it.

vsetvl only returns 0 if the input is 0 so there's no need to test
for 0 after the vsetvl. It does add more lines to C code, but generates
shorter and probably faster assembly.

I also replaced some ints with size_t to avoid sext.w in the generated assembly.

### DIFF
--- a/rvv_saxpy.c
+++ b/rvv_saxpy.c
@@ -54,7 +54,8 @@ void saxpy_vec(size_t n, const float a, const float *x, float *y) {
 
   vfloat32m8_t vx, vy;
 
-  for (; (l = vsetvl_e32m8(n)) > 0; n -= l) {
+  for (; n > 0; n -= l) {
+    l = vsetvl_e32m8(n);
     vx = vle32_v_f32m8(x, l);
     x += l;
     vy = vle32_v_f32m8(y, l);
--- a/rvv_sgemm.c
+++ b/rvv_sgemm.c
@@ -68,10 +68,11 @@ void sgemm_vec(size_t size_m, size_t size_n, size_t size_k,
                float *c, // m * n matrix
                size_t ldc) {
   size_t vl;
-  for (int m = 0; m < size_m; ++m) {
+  for (size_t m = 0; m < size_m; ++m) {
     const float *b_n_ptr = b;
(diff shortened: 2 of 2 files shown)


aswaterman:
I certainly approve of the `size_t` change.  IIRC, @nick-knight prefers the existing code for the loop structure, though.

nick-knight:
My preference for the previous version is purely pedagogical and aesthetic. I agree that the current version will avoid one unnecessary setvl (overall, not per-iteration), and I use that form in my own work. 

One question I have is, why can't the compiler rewrite the current version as the proposed? I guess the answer is that setvl has side effects, even when AVL = 0. I'm wondering if these side effects are something that are (reliably) exposed to the C programmer elsewhere, given the leeway the compiler has in inserting/hoisting setvls elsewhere. For example, if I read `vtype` (via inline asm, I think the only option?) are there any guarantees I will see a certain value given the neighboring intrinsics?

topperc:
I think the compiler could rewrite it, as you said the vtype and vl register side effects aren't reliably visible.

The compiler could rewrite the compares vsetvl output and 0 to use the vsetvl input instead. Then look for a phi node with all its inputs from vsetvls with identical vtypes that have no other uses. We then pull the vsetvli through the phi and rewrite the phi to use the AVL input from the original vsetvls. Finally we need to CSE the new phi node with the other phi node that probably exists for the induction variable. I think that's all pretty doable.

ebahapo:
I don't think that there's any pedagogical difference between the previous and new code for the loop WRT how to program for RVV.  Rather, the new code has the pedagogical advantage WRT how to program for RVV optimally and avoiding a pitfall.

zakk0610:
LGTM. 
IMO, I like the new one because it could be faster.

What do you think? @Hsiangkai @kito-cheng

Hsiangkai:
I am fine with the new one.

nick-knight:
I hope I'm not holding up this PR: I'm the only one who raised any objection, and @ebahapo's comment changed my mind (regarding pedagogy).

PR #97, opened 2021-06-08
Add notes that reductions, vmv.s.x, and vfmv.s.f will use tail undist…

topperc:
…urbed policy unless the dest operand is vundefined().


topperc:
The behavior changed between 0.9 and 0.10. These instructions now obey tail policy instead of just writing element 0. I missed this in the compiler implementation.

Hsiangkai (approved):
LGTM.

Issue #95, opened 2021-06-01
Proposal for alternate interface for segment load/store and tuple types.


topperc:
We've had some implementation issues supporting the tuple types in clang and LLVM IR. This is due to lack of support for structures of scalable vectors in LLVM IR. AArch64 SVE was able to work around this limitation, but the existence of LMUL for RISCV has made it impossible to adopt their approach.

We are able to represent the load/store intrinsics in IR without issue, but the problems show up when we need to store the tuple to a local variable to implement the assignment operator for tuple types. Or when we need to load the tuple from the local variable to use it.

To avoid these implemention issues we at SiFive would like to propose a new interface that avoids the tuples and removes  the `vget`, `vset`, `vcreate`, and `vcopy` intrinsics. We're looking for feedback on this.

For segment stores, we propose to pass between 2 and 8 vectors as separate arguments to the intrinsic rather than a tuple. This combines `vcreate` and store into a single intrinsic. For example

```
void vsseg2e8_v_i8mf8 (int8_t *base, vint8mf8_t v0, vint8mf8_t v1, size_t vl);
void vsseg8e8_v_i8mf8_m (vbool64_t mask, int8_t *base, vint8mf8_t v0, vint8mf8_t v1, vint8mf8_t v2, vint8mf8_t v3, vint8mf8_t v4, vint8mf8_t v5, vint8mf8_t v6, vint8mf8_t v7, size_t vl);
void vssseg2e8_v_i8mf8 (int8_t *base, ptrdiff_t bstride, vint8mf8_t v0, vint8mf8_t v1, size_t vl);
void vssseg8e8_v_i8mf8_m (vbool64_t mask, int8_t *base, ptrdiff_t bstride, vint8mf8_t v0, vint8mf8_t v1, vint8mf8_t v2, vint8mf8_t v3, vint8mf8_t v4, vint8mf8_t v5, vint8mf8_t v6, vint8mf8_t v7, size_t vl);
```

Loads are more complicated because we can't return multiple values in C and we can't make a struct of scalable vectors in C either. We propose to make the intrinsics return void and pass the address of local variables to store the results in 2-8 pointer arguments. These pointers cannot be null. Maskedoff values will be passed as 2-8 additonal arguments. For example:

```
void vlseg2e8_v_i8mf8 (vint8mf8_t *v0, vint8mf8_t *v1, const int8_t *base, size_t vl);
[...]

PR #98, opened 2021-06-08
New interface for segment load/store and remove tuple types.

kito-cheng:
This PR is for https://github.com/riscv/rvv-intrinsic-doc/issues/95.

- New interface for segment load/store
- Remove all tuple type related API


nick-knight (changes requested):
Perhaps I'm misunderstanding this commit, but there appear to be some changes unrelated to Zvlsseg and tuple types. For example, the semantic vector functions are deleted. I suggest splitting this into multiple PRs.

kito-cheng:
Thanks, added back :)

Hsiangkai:
LGTM.

topperc:
Are we ready to push this?

Hsiangkai:
There are conflicting files. @kito-cheng may need to update the PR based on the latest version.

zakk0610:
> Are we ready to push this?

Yes, #95 have more people agree this changed, so LGTM.

Hsiangkai (approved):
Let's get it in to have a workable solution first. If there is better solution later, we could refine it.

We have different intrinsic name, `vlseg2e8_v_i8m1`, from the tuple one, `vlseg2e8_v_i8m1x2`. If we want to recover tuple types later, there is no conflict here.

kito-cheng:
Rebased.

Hsiangkai:
Why are semantic vector functions deleted?

kito-cheng:
Add back semantic vector functions.

Hsiangkai:
LGTM.

Issue #27, opened 2020-06-23
Should intrinsic have to model tail elements?


zakk0610:
current intrinsic RFC does not model tail element, 
but someone maybe want to set the tail undisturbed like https://github.com/riscv/riscv-v-spec/issues/157#issuecomment-527104675.
I think maybe we can only provide additional tail operand for reduction intrinsic functions to model tail elements.
any idea?

knightsifive:
My opinion is that all functionality exposed to the RVV assembly programmer should also be exposed to the RVV intrinsics programmer; this includes [the `vta` and `vma` settings](https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc#333-vector-tail-agnostic-and-vector-mask-agnostic-vta-and-vma). Since these bits are part of `vtype` (along with `vsew` and `vlmul`), it raises the question of whether they should also be encoded into the intrinsic names, passed in as arguments, etc.

rdolbeau:
My 2 cents on the subject...

In most case when coding at high level, unused values in the registers (both inactive and tail) are irrelevant and the user doesn't care [1] what's in it. For peel/tail they are not loaded, not computed and not stored. For masked-out values (e.g. convergence loops), it's the same. The one obvious big exception is conditionals in loop, where usually the old value should be kept (either because the condition if false for 'if/then', or we're in the 'else' merging with the 'then' for 'if/then/else').

So in my opinion the default behavior should be 'keep whatever is in there' (undisturbed), as it should be quite enough for most use cases (as it satisfies both 'don't care' and 'merge' semantic). That's also why vundefined() is in there - it explicits the 'don't care' by inputing undefined value in the 'merge' semantic. For user in need of a specific value instead of undefined(), they can easily use an explicit intrinsic or a value as an input.

So in v0.9-speak, default current behavior should be either {mu, tu} (if the maskedoff input is valid) or {ma,ta} (if it's vundefined() or not present in the intrinsics, thus implicitly undefined, e.g. the unmasked intrinsics).

The question is, how to also model {ma,tu} and {mu,ta} in the intrinsics... and how does that interact with the above? In the current design, having {*,tu} or {mu,*} only works if there is a valid maskedoff parameter, otherwise the user can't specify the values to be left undisturbed... but from my reading of v0.9, the non-masked variant should be eligible for {*,tu} (not that {mu,tu} makes much sense w/o a mask...).

[...]

PR #101, opened 2021-06-17
Add an argument for tail agnostic/undisturbed.

Hsiangkai:
This is a draft version for the intrinsics with the tail agnostic/undisturbed argument. This patch is a preliminary work for #27.

**Suffix:**
with mask and tail policy: _mt
with mask only: _m
with tail only: _t

=====
In most cases, add `_mt` intrinsics with `ta` argument to control the tail policy.

**Use `dest` argument to control tail policy**
* vcompress
* vmv.s.x
* vfmv.s.f
* vmerge.vvm/vmerge.vxm/vmerge.vim (add `_t` intrinsics with additional `dest` argument)
* vfmerge.vfm (add `_t` intrinsics with additional `dest` argument)
* vmv.v.v (add `_t` intrinsics with additional `dest` argument)
* vmv.v.x (add `_t` intrinsics with additional `dest` argument)
* vmv.v.i (add `_t` intrinsics with additional `dest` argument)
* vfmv.v.f (add `_t` intrinsics with additional `dest` argument)
* vadc.vvm/vadc.vxm/vadc.vim (add `_t` intrinsics with additional `dest` argument)
* vsbc.vvm/vsbc.vxm (add `_t` intrinsics with additional `dest` argument)

**Always has tail argument for masked/unmasked intrinsics**
* Vector Single-Width Integer Multiply-Add Instructions (add `_t` and `_mt` intrinsics)
* Vector Widening Integer Multiply-Add Instructions (add `_t` and `_mt` intrinsics)
* Vector Single-Width Floating-Point Fused Multiply-Add Instructions (add `_t` and `_mt` intrinsics)
* Vector Widening Floating-Point Fused Multiply-Add Instructions (add `_t` and `_mt` intrinsics)
* Vector Reduction Operations (add `_t` and `_mt` intrinsics)
* Vector Slideup Instructions (add `_t` and `_mt` intrinsics)
* Vector Slidedown Instructions (add `_t` and `_mt` intrinsics)

**Always tail-agnostic policy (No need to add `ta` argument)**
* Vector Integer Comparison Instructions
* Vector Floating-Point Compare Instructions
* Vector Floating-Point Classify Instruction
* vmadc.vvm/vmadc.vxm/vmadc.vim
* vmadc.vv/vmadc.vx/vmadc.vi
* vmsbc.vvm/vmsbc.vxm
* vmsbc.vv/vmsbc.vx
* vmsbf.m
* vmsif.m
* vmsof.m
* vpopc.m
* vmv.x.s
* vfmv.f.s


topperc:
I wonder if the polarity of the operand should match the polarity of the bit in the vsetvli encoding. So the bit would be ta rather than tu.

overloaded_intrinsic_funcs/09_vector_mask_functions.md
```diff
@@ -202,13 +202,13 @@ vbool16_t vmsof (vbool16_t op1, size_t vl);
...
 vbool64_t vmsof (vbool64_t op1, size_t vl);
 // masked functions
-vbool1_t vmsof (vbool1_t mask, vbool1_t maskedoff, vbool1_t op1, size_t vl);
-vbool2_t vmsof (vbool2_t mask, vbool2_t maskedoff, vbool2_t op1, size_t vl);
-vbool4_t vmsof (vbool4_t mask, vbool4_t maskedoff, vbool4_t op1, size_t vl);
-vbool8_t vmsof (vbool8_t mask, vbool8_t maskedoff, vbool8_t op1, size_t vl);
-vbool16_t vmsof (vbool16_t mask, vbool16_t maskedoff, vbool16_t op1, size_t vl);
-vbool32_t vmsof (vbool32_t mask, vbool32_t maskedoff, vbool32_t op1, size_t vl);
-vbool64_t vmsof (vbool64_t mask, vbool64_t maskedoff, vbool64_t op1, size_t vl);
+vbool1_t vmsof (vbool1_t mask, vbool1_t maskedoff, vbool1_t op1, size_t vl, uint8_t tu);
```
topperc, line 205: Isn't this instruction always tail agnostic? It's a mask destination

overloaded_intrinsic_funcs/09_vector_mask_functions.md
```diff
@@ -182,13 +182,13 @@ vbool16_t vmsif (vbool16_t op1, size_t vl);
...
 vbool64_t vmsif (vbool64_t op1, size_t vl);
 // masked functions
-vbool1_t vmsif (vbool1_t mask, vbool1_t maskedoff, vbool1_t op1, size_t vl);
-vbool2_t vmsif (vbool2_t mask, vbool2_t maskedoff, vbool2_t op1, size_t vl);
-vbool4_t vmsif (vbool4_t mask, vbool4_t maskedoff, vbool4_t op1, size_t vl);
-vbool8_t vmsif (vbool8_t mask, vbool8_t maskedoff, vbool8_t op1, size_t vl);
-vbool16_t vmsif (vbool16_t mask, vbool16_t maskedoff, vbool16_t op1, size_t vl);
-vbool32_t vmsif (vbool32_t mask, vbool32_t maskedoff, vbool32_t op1, size_t vl);
-vbool64_t vmsif (vbool64_t mask, vbool64_t maskedoff, vbool64_t op1, size_t vl);
+vbool1_t vmsif (vbool1_t mask, vbool1_t maskedoff, vbool1_t op1, size_t vl, uint8_t tu);
```
topperc, line 185: Isn't this instruction always tail agnostic? It's a mask destination

overloaded_intrinsic_funcs/09_vector_mask_functions.md
```diff
@@ -162,13 +162,13 @@ vbool16_t vmsbf (vbool16_t op1, size_t vl);
...
 vbool64_t vmsbf (vbool64_t op1, size_t vl);
 // masked functions
-vbool1_t vmsbf (vbool1_t mask, vbool1_t maskedoff, vbool1_t op1, size_t vl);
-vbool2_t vmsbf (vbool2_t mask, vbool2_t maskedoff, vbool2_t op1, size_t vl);
-vbool4_t vmsbf (vbool4_t mask, vbool4_t maskedoff, vbool4_t op1, size_t vl);
-vbool8_t vmsbf (vbool8_t mask, vbool8_t maskedoff, vbool8_t op1, size_t vl);
-vbool16_t vmsbf (vbool16_t mask, vbool16_t maskedoff, vbool16_t op1, size_t vl);
-vbool32_t vmsbf (vbool32_t mask, vbool32_t maskedoff, vbool32_t op1, size_t vl);
-vbool64_t vmsbf (vbool64_t mask, vbool64_t maskedoff, vbool64_t op1, size_t vl);
+vbool1_t vmsbf (vbool1_t mask, vbool1_t maskedoff, vbool1_t op1, size_t vl, uint8_t tu);
```
topperc, line 165: Isn't this instruction always tail agnostic? It's a mask destination

overloaded_intrinsic_funcs/09_vector_mask_functions.md
```diff
@@ -6,13 +6,13 @@
...
 **Prototypes:**
 ``` C
-vbool1_t vlm_v_b1 (const uint8_t *base, size_t vl);
-vbool2_t vlm_v_b2 (const uint8_t *base, size_t vl);
-vbool4_t vlm_v_b4 (const uint8_t *base, size_t vl);
-vbool8_t vlm_v_b8 (const uint8_t *base, size_t vl);
-vbool16_t vlm_v_b16 (const uint8_t *base, size_t vl);
-vbool32_t vlm_v_b32 (const uint8_t *base, size_t vl);
-vbool64_t vlm_v_b64 (const uint8_t *base, size_t vl);
+vbool1_t vlm (const uint8_t *base, size_t vl);
```
topperc, line 9: This function name is ambiguous now isn't it?

overloaded_intrinsic_funcs/10_vector_permutation_functions.md
```diff
@@ -7,365 +7,365 @@
 **Prototypes:**
 ``` C
 int8_t vmv_x (vint8mf8_t src);
-vint8mf8_t vmv_s (vint8mf8_t dest, int8_t src, size_t vl);
+vint8mf8_t vmv_s (vint8mf8_t dest, int8_t src, size_t vl, uint8_t tu);
```
topperc, line 10: Isn't this instruction's tail policy already covered by dest being vundefined or not?

overloaded_intrinsic_funcs/10_vector_permutation_functions.md
```diff
@@ -7,365 +7,365 @@
...
-vfloat32mf2_t vslidedown (vfloat32mf2_t dest, vfloat32mf2_t src, size_t offset, size_t vl);
-vfloat32m1_t vslidedown (vfloat32m1_t dest, vfloat32m1_t src, size_t offset, size_t vl);
-vfloat32m2_t vslidedown (vfloat32m2_t dest, vfloat32m2_t src, size_t offset, size_t vl);
-vfloat32m4_t vslidedown (vfloat32m4_t dest, vfloat32m4_t src, size_t offset, size_t vl);
-vfloat32m8_t vslidedown (vfloat32m8_t dest, vfloat32m8_t src, size_t offset, size_t vl);
-vfloat64m1_t vslidedown (vfloat64m1_t dest, vfloat64m1_t src, size_t offset, size_t vl);
-vfloat64m2_t vslidedown (vfloat64m2_t dest, vfloat64m2_t src, size_t offset, size_t vl);
-vfloat64m4_t vslidedown (vfloat64m4_t dest, vfloat64m4_t src, size_t offset, size_t vl);
-vfloat64m8_t vslidedown (vfloat64m8_t dest, vfloat64m8_t src, size_t offset, size_t vl);
+vint8mf8_t vslideup (vint8mf8_t dest, vint8mf8_t src, size_t offset, size_t vl, uint8_t tu);
```
topperc, line 132: Wasn't this already handled by dest being vundefined?

overloaded_intrinsic_funcs/10_vector_permutation_functions.md
```diff
@@ -789,243 +789,243 @@ vfloat64m2_t vrgatherei16 (vfloat64m2_t op1, vuint16mf2_t op2, size_t vl);
...
-vfloat32mf2_t vcompress (vbool64_t mask, vfloat32mf2_t dest, vfloat32mf2_t src, size_t vl);
-vfloat32m1_t vcompress (vbool32_t mask, vfloat32m1_t dest, vfloat32m1_t src, size_t vl);
-vfloat32m2_t vcompress (vbool16_t mask, vfloat32m2_t dest, vfloat32m2_t src, size_t vl);
-vfloat32m4_t vcompress (vbool8_t mask, vfloat32m4_t dest, vfloat32m4_t src, size_t vl);
-vfloat32m8_t vcompress (vbool4_t mask, vfloat32m8_t dest, vfloat32m8_t src, size_t vl);
-vfloat64m1_t vcompress (vbool64_t mask, vfloat64m1_t dest, vfloat64m1_t src, size_t vl);
-vfloat64m2_t vcompress (vbool32_t mask, vfloat64m2_t dest, vfloat64m2_t src, size_t vl);
-vfloat64m4_t vcompress (vbool16_t mask, vfloat64m4_t dest, vfloat64m4_t src, size_t vl);
-vfloat64m8_t vcompress (vbool8_t mask, vfloat64m8_t dest, vfloat64m8_t src, size_t vl);
+vint8mf8_t vcompress (vbool64_t mask, vint8mf8_t dest, vint8mf8_t src, size_t vl, uint8_t tu);
```
topperc, line 972: Wasn't this already handled by dest being vundefined?

overloaded_intrinsic_funcs/08_vector_reduction_functions.md
```diff
@@ -6,787 +6,787 @@
...
-vuint32m1_t vredxor (vuint32m1_t dest, vuint32mf2_t vector, vuint32m1_t scalar, size_t vl);
-vuint32m1_t vredxor (vuint32m1_t dest, vuint32m1_t vector, vuint32m1_t scalar, size_t vl);
-vuint32m1_t vredxor (vuint32m1_t dest, vuint32m2_t vector, vuint32m1_t scalar, size_t vl);
-vuint32m1_t vredxor (vuint32m1_t dest, vuint32m4_t vector, vuint32m1_t scalar, size_t vl);
-vuint32m1_t vredxor (vuint32m1_t dest, vuint32m8_t vector, vuint32m1_t scalar, size_t vl);
-vuint64m1_t vredxor (vuint64m1_t dest, vuint64m1_t vector, vuint64m1_t scalar, size_t vl);
-vuint64m1_t vredxor (vuint64m1_t dest, vuint64m2_t vector, vuint64m1_t scalar, size_t vl);
-vuint64m1_t vredxor (vuint64m1_t dest, vuint64m4_t vector, vuint64m1_t scalar, size_t vl);
-vuint64m1_t vredxor (vuint64m1_t dest, vuint64m8_t vector, vuint64m1_t scalar, size_t vl);
+vint8m1_t vredsum (vint8m1_t dest, vint8mf8_t vector, vint8m1_t scalar, size_t vl, uint8_t tu);
```
topperc, line 9: Wasn't this already handled by dest being vundefined?

overloaded_intrinsic_funcs/07_vector_floating-point_functions.md
```diff
@@ -1813,186 +1813,186 @@ vbool16_t vmfge (vfloat64m4_t op1, float64_t op2, size_t vl);
...
-vbool4_t vmfge (vbool4_t mask, vbool4_t maskedoff, vfloat32m8_t op1, float32_t op2, size_t vl);
-vbool64_t vmfge (vbool64_t mask, vbool64_t maskedoff, vfloat64m1_t op1, vfloat64m1_t op2, size_t vl);
-vbool64_t vmfge (vbool64_t mask, vbool64_t maskedoff, vfloat64m1_t op1, float64_t op2, size_t vl);
-vbool32_t vmfge (vbool32_t mask, vbool32_t maskedoff, vfloat64m2_t op1, vfloat64m2_t op2, size_t vl);
-vbool32_t vmfge (vbool32_t mask, vbool32_t maskedoff, vfloat64m2_t op1, float64_t op2, size_t vl);
-vbool16_t vmfge (vbool16_t mask, vbool16_t maskedoff, vfloat64m4_t op1, vfloat64m4_t op2, size_t vl);
-vbool16_t vmfge (vbool16_t mask, vbool16_t maskedoff, vfloat64m4_t op1, float64_t op2, size_t vl);
-vbool8_t vmfge (vbool8_t mask, vbool8_t maskedoff, vfloat64m8_t op1, vfloat64m8_t op2, size_t vl);
-vbool8_t vmfge (vbool8_t mask, vbool8_t maskedoff, vfloat64m8_t op1, float64_t op2, size_t vl);
+vbool64_t vmfeq (vbool64_t mask, vbool64_t maskedoff, vfloat16mf4_t op1, vfloat16mf4_t op2, size_t vl, uint8_t tu);
```
topperc, line 1816: Compares should always be tail agnostic since they have a mask destination.

topperc (commented):
Do you plan to add tail undisturbed intrinsics that provide a dest operand for

vmv.v.v/vmv.v.x/vmv.v.i
vmerge.vvm/vmerge.vxm/vmerge.vim/vfmerge.vfm 
vadc.vvm/vadc.vxm/vadc.vim
vsbc.vvm/vsbc.vxm

intrinsic_funcs/05_vector_integer_arithmetic_functions.md
```diff
@@ -3274,534 +3274,534 @@ vbool16_t vmsgeu_vx_u64m4_b16 (vuint64m4_t op1, uint64_t op2, size_t vl);
...
-vbool4_t vmsgeu_vx_u32m8_b4_m (vbool4_t mask, vbool4_t maskedoff, vuint32m8_t op1, uint32_t op2, size_t vl);
-vbool64_t vmsgeu_vv_u64m1_b64_m (vbool64_t mask, vbool64_t maskedoff, vuint64m1_t op1, vuint64m1_t op2, size_t vl);
-vbool64_t vmsgeu_vx_u64m1_b64_m (vbool64_t mask, vbool64_t maskedoff, vuint64m1_t op1, uint64_t op2, size_t vl);
-vbool32_t vmsgeu_vv_u64m2_b32_m (vbool32_t mask, vbool32_t maskedoff, vuint64m2_t op1, vuint64m2_t op2, size_t vl);
-vbool32_t vmsgeu_vx_u64m2_b32_m (vbool32_t mask, vbool32_t maskedoff, vuint64m2_t op1, uint64_t op2, size_t vl);
-vbool16_t vmsgeu_vv_u64m4_b16_m (vbool16_t mask, vbool16_t maskedoff, vuint64m4_t op1, vuint64m4_t op2, size_t vl);
-vbool16_t vmsgeu_vx_u64m4_b16_m (vbool16_t mask, vbool16_t maskedoff, vuint64m4_t op1, uint64_t op2, size_t vl);
-vbool8_t vmsgeu_vv_u64m8_b8_m (vbool8_t mask, vbool8_t maskedoff, vuint64m8_t op1, vuint64m8_t op2, size_t vl);
-vbool8_t vmsgeu_vx_u64m8_b8_m (vbool8_t mask, vbool8_t maskedoff, vuint64m8_t op1, uint64_t op2, size_t vl);
+vbool64_t vmseq_vv_i8mf8_b64_m (vbool64_t mask, vbool64_t maskedoff, vint8mf8_t op1, vint8mf8_t op2, size_t vl, uint8_t ta);
```
topperc, line 3277: These should always be tail agnostic

intrinsic_funcs/05_vector_integer_arithmetic_functions.md
```diff
@@ -5060,1338 +5060,1338 @@ vint64m4_t vwmulsu_vx_i64m4 (vint32m2_t op1, uint32_t op2, size_t vl);
...
-vuint32m8_t vnmsub_vx_u32m8 (vuint32m8_t acc, uint32_t op1, vuint32m8_t op2, size_t vl);
-vuint64m1_t vnmsub_vv_u64m1 (vuint64m1_t acc, vuint64m1_t op1, vuint64m1_t op2, size_t vl);
-vuint64m1_t vnmsub_vx_u64m1 (vuint64m1_t acc, uint64_t op1, vuint64m1_t op2, size_t vl);
-vuint64m2_t vnmsub_vv_u64m2 (vuint64m2_t acc, vuint64m2_t op1, vuint64m2_t op2, size_t vl);
-vuint64m2_t vnmsub_vx_u64m2 (vuint64m2_t acc, uint64_t op1, vuint64m2_t op2, size_t vl);
-vuint64m4_t vnmsub_vv_u64m4 (vuint64m4_t acc, vuint64m4_t op1, vuint64m4_t op2, size_t vl);
-vuint64m4_t vnmsub_vx_u64m4 (vuint64m4_t acc, uint64_t op1, vuint64m4_t op2, size_t vl);
-vuint64m8_t vnmsub_vv_u64m8 (vuint64m8_t acc, vuint64m8_t op1, vuint64m8_t op2, size_t vl);
-vuint64m8_t vnmsub_vx_u64m8 (vuint64m8_t acc, uint64_t op1, vuint64m8_t op2, size_t vl);
+vint8mf8_t vmacc_vv_i8mf8 (vint8mf8_t acc, vint8mf8_t op1, vint8mf8_t op2, size_t vl, uint8_t tu);
```
topperc, line 5158: This still says tu

intrinsic_funcs/05_vector_integer_arithmetic_functions.md
```diff
@@ -5060,1338 +5060,1338 @@ vint64m4_t vwmulsu_vx_i64m4 (vint32m2_t op1, uint32_t op2, size_t vl);
...
-vint32mf2_t vqmaccus_vx_i32mf2 (vint32mf2_t acc, uint8_t op1, vint8mf8_t op2, size_t vl);
-vint32m1_t vqmaccus_vx_i32m1 (vint32m1_t acc, uint8_t op1, vint8mf4_t op2, size_t vl);
-vint32m2_t vqmaccus_vx_i32m2 (vint32m2_t acc, uint8_t op1, vint8mf2_t op2, size_t vl);
-vint32m4_t vqmaccus_vx_i32m4 (vint32m4_t acc, uint8_t op1, vint8m1_t op2, size_t vl);
-vint32m8_t vqmaccus_vx_i32m8 (vint32m8_t acc, uint8_t op1, vint8m2_t op2, size_t vl);
-vint64m1_t vqmaccus_vx_i64m1 (vint64m1_t acc, uint16_t op1, vint16mf4_t op2, size_t vl);
-vint64m2_t vqmaccus_vx_i64m2 (vint64m2_t acc, uint16_t op1, vint16mf2_t op2, size_t vl);
-vint64m4_t vqmaccus_vx_i64m4 (vint64m4_t acc, uint16_t op1, vint16m1_t op2, size_t vl);
-vint64m8_t vqmaccus_vx_i64m8 (vint64m8_t acc, uint16_t op1, vint16m2_t op2, size_t vl);
+vint32mf2_t vqmacc_vv_i32mf2 (vint32mf2_t acc, vint8mf8_t op1, vint8mf8_t op2, size_t vl, uint8_t tu);
```
topperc, line 6084: Not related to this patch, but should quad widening intrinsics be removed since they aren't in the V spec drafts

intrinsic_funcs/08_vector_reduction_functions.md
```diff
@@ -271,270 +271,270 @@ vuint64m1_t vredxor_vs_u64m2_u64m1 (vuint64m1_t dest, vuint64m2_t vector, vuint6
...
-vuint32m1_t vredxor_vs_u32m1_u32m1_m (vbool32_t mask, vuint32m1_t dest, vuint32m1_t vector, vuint32m1_t scalar, size_t vl);
-vuint32m1_t vredxor_vs_u32m2_u32m1_m (vbool16_t mask, vuint32m1_t dest, vuint32m2_t vector, vuint32m1_t scalar, size_t vl);
-vuint32m1_t vredxor_vs_u32m4_u32m1_m (vbool8_t mask, vuint32m1_t dest, vuint32m4_t vector, vuint32m1_t scalar, size_t vl);
-vuint32m1_t vredxor_vs_u32m8_u32m1_m (vbool4_t mask, vuint32m1_t dest, vuint32m8_t vector, vuint32m1_t scalar, size_t vl);
-vuint64m1_t vredxor_vs_u64m1_u64m1_m (vbool64_t mask, vuint64m1_t dest, vuint64m1_t vector, vuint64m1_t scalar, size_t vl);
-vuint64m1_t vredxor_vs_u64m2_u64m1_m (vbool32_t mask, vuint64m1_t dest, vuint64m2_t vector, vuint64m1_t scalar, size_t vl);
-vuint64m1_t vredxor_vs_u64m4_u64m1_m (vbool16_t mask, vuint64m1_t dest, vuint64m4_t vector, vuint64m1_t scalar, size_t vl);
-vuint64m1_t vredxor_vs_u64m8_u64m1_m (vbool8_t mask, vuint64m1_t dest, vuint64m8_t vector, vuint64m1_t scalar, size_t vl);
+vint8m1_t vredsum_vs_i8mf8_i8m1_m (vbool64_t mask, vint8m1_t dest, vint8mf8_t vector, vint8m1_t scalar, size_t vl, uint8_t ta);
```
topperc, line 274: I think reductions can use dest==vundefined for ta. The mask doesn't affect which elements are written in the destination.

kito-cheng:
One idea for this API change, C don't have default argument, but we could use overload trick to make it available.

topperc:
> How about keep the original version, like default argument for ta argument.
> C don't have default argument, but we could use overload trick to make it available.

The compiler currently uses TU, so we'd have to default to that to be compatible, but that's wrong for most use cases. So I'm not sure having a default is an advantage.

kito-cheng:
> The compiler currently uses TU, so we'd have to default to that to be compatible, but that's wrong for most use cases. So I'm not sure having a default is an advantage other than to avoid software churn.

Current intrinsic said we are using tail agnostic, but I think TU is safe for the semantic. 
```
There is no need to specify the behavior of tail and masked-off elements being undisturbed or agnostic. The default setting is tail agnostic and masked-off undisturbed. If users do not want to keep the values in masked-off elements, they could pass vundefined() as the maskedoff value.
```

My exception is most people don't care about `tail`, so default to `tail agnostic` should be safe, in case people want `tail undisturbed` they could use extra argument to specify that, but I am not kind of really intrinsic user, @nick-knight @rdolbeau @rofirrim @rednoah91 @lazyparser @rjiejie could you guys to comment this?


And I would prefer add new function rather than disturb existing API, the reason is same, I expect user don't care in most scenario, if so, why we need force user always specify that.

topperc:
> > The compiler currently uses TU, so we'd have to default to that to be compatible, but that's wrong for most use cases. So I'm not sure having a default is an advantage other than to avoid software churn.
> 
> Current intrinsic said we are using tail agnostic, but I think TU is safe for the semantic.
> 
> ```
> There is no need to specify the behavior of tail and masked-off elements being undisturbed or agnostic. The default setting is tail agnostic and masked-off undisturbed. If users do not want to keep the values in masked-off elements, they could pass vundefined() as the maskedoff value.
> ```
> 
> My exception is most people don't care about `tail`, so default to `tail agnostic` should be safe, in case people want `tail undisturbed` they could use extra argument to specify that, but I am not kind of really intrinsic user, @nick-knight @rdolbeau @rofirrim @rednoah91 @lazyparser @rjiejie could you guys to comment this?
> 
> And I would prefer add new function rather than disturb existing API, the reason is same, I expect user don't care in most scenario, if so, why we need force user always specify that.

The documentation says we default to tail agnostic, but that's not what has been implemented. We know there is code that uses an all ones mask with masked intrinsics to get tail undisturbed or code that has relied on the tail undisturbed behavior of the FMA intrinsics. I don't think we can change the existing intrinsics to use tail agnostic as that would silently miscompile such code without feedback to the user. The code would continue to work until a future out of order core that implements tail agnostic differently than tail undisturbed.

Therefore, I propose either adding a new argument to the masked and fma intrinsics to specify tail policy or renaming them to include `_tu` in their name and adding new intrinsics with `_ta` in their name. This way any code that uses them will need to be explicitly updated to tell the compiler the behavior that is expected.

For intrinsics that don't already have a dest operand, maybe it is best to add a new _tu version with the dest operand and the leave the existing intrinsics alone.

rofirrim:
I'd too have preferred for the implementation to default to `ta` regardless of the mask (this is what `mu` should be for after all, if there is a maskedoff). I understand tail agnostic as defining less  behaviour than tail undisturbed (notwithstanding what the implementation will do) and so it would fit existing intrinsics that don't specify a destination value for the tail elements.

However I understand the slightly redundancy we have when the current intrinsics do receive a maskedoff operand (I'd advocate for `ta,mu` here) and the fact that people have used them with the expectation of `tu,mu`.

I suggest we give some freedom to the implementation to choose `ta`/`tu` but we are explicit in saying that the tail is not necessarily preserved (long vector machines will favour `ta` possibly via some `-mcpu` flag or the likes)

Then we allow the current intrinsics to specify either case desired (`tu` or `ta`) so we give maximum control to the user (which I understand would align with some of @clattner comment).

What do you think?

topperc:
> I too would have preferred for the implementation to default to `ta` regardless of the mask (this is what `mu` should be for after all, if there is a maskedoff). I understand tail agnostic as defining less behaviour than tail undisturbed (notwithstanding what the implementation will do) and so it would fit existing intrinsics that don't specify a destination value for the tail elements.
> 
> However I understand the slightly redundancy we have when the current intrinsics do receive a maskedoff operand (I'd advocate for `ta,mu` here) and the fact that people have used them with the expectation of `tu,mu`.
> 
> I suggest for the existing intrinsics (not the new ones being added) we give some freedom to the implementation to choose `ta`/`tu` but we are explicit in saying that the tail is not necessarily preserved (long vector machines will favour `ta` possibly via some `-mcpu` flag or the likes)

Are you ok with changing the names or number of arguments to the masked intrinsics? Or do you want to change them to ta and just add new versions meaning we may miscompile existing code?

> 
> Then we allow the new intrinsics to specify either case desired (`tu` or `ta`) so we give maximum control to the user (which I understand would align with some of @clattner comment).
> 
> What do you think?

rofirrim:
> Are you ok with changing the names or number of arguments to the masked intrinsics? Or do you want to change them to ta and just add new versions meaning we may miscompile existing code that was relying on tu,mu?

Sorry if it feels we're going circles here.

I like your proposal here https://github.com/riscv/rvv-intrinsic-doc/issues/27#issuecomment-849148244 but with a minor change.

Name the new ones different, for instance `m_t`.

```
vint64m1_t vadd_vv_i64m1_m_t (vbool64_t mask,
    vint64m1_t undisturbed, // the tail of this operand is only considered if tail_policy == TAIL_UNDISTURBED
    vint64m1_t op1, vint64m1_t op2, size_t vl,
    uint8_t tail_policy);
```

And then make the original ones default to `ta` (e.g. via macros). Yes, this would miscompile some existing codes that assumed that the tail of the maskedoff value had defiend behaviour in the result of the intrinsic.

In fact the function above is now very general and perhaps tail_policy could be a bitmap that also allows specifying `MASK_AGNOSTIC` and `MASK_UNDISTURBED` where only the latter gives defined meaning to the masked off elements of the result.

Something like this:

```
vint64m1_t vadd_vv_i64m1_m_t (vbool64_t mask,
    vint64m1_t undisturbed,
      // - elements of the 'undisturbed' parameter past vl
      //   are only defined 'policy' includes TAIL_UNDISTURBED
      // - elements of the 'undisturbed' parameter that would be masked off
      //   by the 'mask' parameter are only defined if 'policy' includes MASK_UNDISTURBED
    vint64m1_t op1, vint64m1_t op2, size_t vl,
    uint8_t policy);
```

Sorry that I might be broadening the scope a bit but I think your proposal can be made powerful enough to cover all the cases. (You also mentioned a number of ways where we could optimise the mask specially if users have to use `_m_t` forms because they want a tail agnostic policy but their mask is all ones, a `vmset`).

Perhaps you had already considered an approach like this. I understand your concern with possible miscompilations but I believe this has been caused because the semantics of the _tail_ of the maskedoff operand were not fully clear.

Hsiangkai:
> > Are you ok with changing the names or number of arguments to the masked intrinsics? Or do you want to change them to ta and just add new versions meaning we may miscompile existing code that was relying on tu,mu?
> 
> Sorry if it feels we're going circles here.
> 
> I like your proposal here [#27 (comment)](https://github.com/riscv/rvv-intrinsic-doc/issues/27#issuecomment-849148244) but with a minor change.
> 
> Name the new ones different, for instance `m_t`.
If we change the name to `_m_t`, does it mean we add new set of intrinsics? Do we still keep the original `_m` intrinsics? If we remove `_m` intrinsics, it will cause the compilation failed and force the users to use the new intrinsics to specify the policy explicitly.
> 
> ```
> vint64m1_t vadd_vv_i64m1_m_t (vbool64_t mask,
>     vint64m1_t undisturbed, // the tail of this operand is only considered if tail_policy == TAIL_UNDISTURBED
>     vint64m1_t op1, vint64m1_t op2, size_t vl,
>     uint8_t tail_policy);
> ```
> 
> And then make the original ones default to `ta` (e.g. via macros). Yes, this would miscompile some existing codes that assumed that the tail of the maskedoff value had defiend behaviour in the result of the intrinsic.
> 
> In fact the function above is now very general and perhaps tail_policy could be a bitmap that also allows specifying `MASK_AGNOSTIC` and `MASK_UNDISTURBED` where only the latter gives defined meaning to the masked off elements of the result.
> 
> Something like this:
> 
> ```
> vint64m1_t vadd_vv_i64m1_m_t (vbool64_t mask,
>     vint64m1_t undisturbed,
>       // - elements of the 'undisturbed' parameter past vl
>       //   are only defined if 'policy' includes TAIL_UNDISTURBED
>       // - elements of the 'undisturbed' parameter that would be masked off
>       //   by the 'mask' parameter are only defined if 'policy' includes MASK_UNDISTURBED
>     vint64m1_t op1, vint64m1_t op2, size_t vl,
>     uint8_t policy);
> ```
> 
> Sorry that I might be broadening the scope a bit but I think your proposal can be made powerful enough to cover all the cases. (You also mentioned a number of ways where we could optimise the mask specially if users have to use `_m_t` forms because they want a tail agnostic policy but their mask is all ones, a `vmset`).
> 
> Perhaps you had already considered an approach like this. I understand your concern with possible miscompilations but I believe this has been caused because the semantics of the _tail_ of the maskedoff operand were not fully clear.

rofirrim:
> 
> If we change the name to `_m_t`, does it mean we add new set of intrinsics? Do we still keep the original `_m` intrinsics? If we remove `_m` intrinsics, it will cause the compilation failed and force the users to use the new intrinsics to specify the policy explicitly.
> 

I suggest we keep the `_m` intrinsics and we forward them to the `_m_t`intrinsics, hardcoding the `policy` to be `TAIL_AGNOSTIC | MASK_UNDISTURBED`.

I admit I haven't thought very deeply on the implementation cost but I'm assuming @kito-cheng with respect to `riscv_vector.h` mostly removes our existing issue with the large number of intrinsics. (I'm aware there is still a documentation problem with so many intrinsics but I think we can address it via mechanical ways)/

kito-cheng:
> I admit I haven't thought very deeply on the implementation cost but I'm assuming @kito-cheng proposal with respect to riscv_vector.h mostly removes our existing issue with the large number of intrinsics. (I'm aware there is still a documentation problem with so many intrinsics but I think we can address it via mechanical ways)

For the clang/llvm implementation part, I believe large number of intrinsics should not problem soon :)

Hsiangkai:
> > If we change the name to `_m_t`, does it mean we add new set of intrinsics? Do we still keep the original `_m` intrinsics? If we remove `_m` intrinsics, it will cause the compilation failed and force the users to use the new intrinsics to specify the policy explicitly.
> 
> I suggest we keep the `_m` intrinsics and we forward them to the `_m_t`intrinsics, hardcoding the `policy` to be `TAIL_AGNOSTIC | MASK_UNDISTURBED`.

We could use vundefined() on the maskedoff argument to decide mask undisturbed or agnostic. Do we still need to specify it in the additional argument?

> 
> I admit I haven't thought very deeply on the implementation cost but I'm assuming @kito-cheng proposal with respect to `riscv_vector.h` mostly removes our existing issue with the large number of intrinsics. (I'm aware there is still a documentation problem with so many intrinsics but I think we can address it via mechanical ways)

Hsiangkai:
Here is the proof-of-concept patch: https://reviews.llvm.org/D105092.

If all agree to define `_mt` intrinsics for masked operations with tail policy and keep `_m` intrinsics with tail agnostic as default tail policy, I could update the document accordingly.

topperc:
> > I admit I haven't thought very deeply on the implementation cost but I'm assuming @kito-cheng proposal with respect to riscv_vector.h mostly removes our existing issue with the large number of intrinsics. (I'm aware there is still a documentation problem with so many intrinsics but I think we can address it via mechanical ways)
> 
> For the clang/llvm implementation part, I believe large number of intrinsics should not cause compilation time issue soon :)

Does that solution allow macros so that _m can forward to _mt with extra policy argument or would those still have to be in the header file?

rofirrim:
> Here is the proof-of-concept patch: https://reviews.llvm.org/D105092.
> 
> If all agree to define `_mt` intrinsics for masked operations with tail policy and keep `_m` intrinsics with tail agnostic as default tail policy, I could update the document accordingly.
> 
> In this PoC patch, I didn't model `maskedoff == vundefined()` behavior.

Seems reasonable, just double checking: we will be able to lower to an unmasked instruction if the mask is all ones, won't we?

> We could use vundefined() on the maskedoff argument to decide mask undisturbed or agnostic. Do we still need to specify it in the additional argument? My opinion is to use the additional argument for tail elements only.

I'm a bit worried with the `vundefined()` for the maskedoff but I guess it'll allow us to select `ta, ma`, right? In that case I think a proposal where we only specify the tail policy is a balanced trade-off.

kito-cheng:
> Does that solution allow macros so that _m can forward to _mt with extra policy argument or would those still have to be in the header file?

Overload and mapping intrinsic to built-in is OK, but I don't know how to add default argument yet.

Hsiangkai:
I updated the document. I have two questions when modifying the document.
1. Do we need to control tail policy for
* Single-Width Floating-Point/Integer Type-Convert Instructions
* Widening Floating-Point/Integer Type-Convert Instructions
* Narrowing Floating-Point/Integer Type-Convert Instructions

If we keep the tail elements, does it mean parts of the elements are floating point and parts of the elements are integer? Does it make sense to keep tail elements for conversion intrinsics?

2. Some intrinsics use `ta` argument and some intrinsics use `dest == vundefined()?` to control tail policy. Is it acceptable to use different way to control the tail policy? Or we add the `ta` argument even if `dest == vundefined()?` is enough.

It becomes
`ta` == true, we do not care about `dest`'s value.
`ta` == false, and `dest` == vundefined() => optimized to ta,ma.
`ta` == false, and `dest` != vundefined() => use tu,ma

topperc:
> I updated the document. I have two questions when modifying the document.
> 
> 1. Do we need to control tail policy for
> 
> * Single-Width Floating-Point/Integer Type-Convert Instructions
> * Widening Floating-Point/Integer Type-Convert Instructions
> * Narrowing Floating-Point/Integer Type-Convert Instructions
> 
> If we keep the tail elements, does it mean parts of the elements are floating point and parts of the elements are integer? Does it make sense to keep tail elements for conversion intrinsics?

The tail elements would come from a different operand/register right? So they could be integer values if `vd` is a different register than `vs2`.

> 
> 1. Some intrinsics use `ta` argument and some intrinsics use `dest == vundefined()?` to control tail policy. Is it acceptable to use different way to control the tail policy? Or we add the `ta` argument even if `dest == vundefined()?` is enough.
> 
> It becomes
> `ta` == true, we do not care about `dest`'s value.
> `ta` == false, and `dest` == vundefined() => optimized to ta,ma.
> `ta` == false, and `dest` != vundefined() => use tu,ma

If we have the dest operand and ta operand, will the compiler need to replace the dest with undef if agnostic is selected to give freedom to the register allocator and allow the computation to be dead code eliminated if the value passed to dest isn't used anywhere else?

Hsiangkai:
> > I updated the document. I have two questions when modifying the document.
> > 
> > 1. Do we need to control tail policy for
> > 
> > 
> > * Single-Width Floating-Point/Integer Type-Convert Instructions
> > * Widening Floating-Point/Integer Type-Convert Instructions
> > * Narrowing Floating-Point/Integer Type-Convert Instructions
> > 
> > If we keep the tail elements, does it mean parts of the elements are floating point and parts of the elements are integer? Does it make sense to keep tail elements for conversion intrinsics?
> 
> The tail elements would come from a different operand/register right? So they could be integer values if `vd` is a different register than `vs2`.

Added.
> 
> > 1. Some intrinsics use `ta` argument and some intrinsics use `dest == vundefined()?` to control tail policy. Is it acceptable to use different way to control the tail policy? Or we add the `ta` argument even if `dest == vundefined()?` is enough.
> > 
> > It becomes
> > `ta` == true, we do not care about `dest`'s value.
> > `ta` == false, and `dest` == vundefined() => optimized to ta,ma.
> > `ta` == false, and `dest` != vundefined() => use tu,ma
> 
> If we have the dest operand and ta operand, will the compiler need to replace the dest with undef if agnostic is selected to give freedom to the register allocator and allow the computation to be dead code eliminated if the value passed to dest isn't used anywhere else?

I have no strong opinion here. Just curious which way is more "user friendly". Use single argument to control will be easier to implement. I will keep the current version unless someone opposes it.

zakk0610:
> > If we change the name to `_m_t`, does it mean we add new set of intrinsics? Do we still keep the original `_m` intrinsics? If we remove `_m` intrinsics, it will cause the compilation failed and force the users to use the new intrinsics to specify the policy explicitly.
> 
> I suggest we keep the `_m` intrinsics and we forward them to the `_m_t`intrinsics, hardcoding the `policy` to be `TAIL_AGNOSTIC | MASK_UNDISTURBED`.

I would like to keep the `_m` intrinsics and changed the default tail policy to `ta`, because I believe most people don't care about tail.  In additional, v is a experimental extension, so maybe it's fine to having the backward compatibility problem.

> Some intrinsics use ta argument and some intrinsics use dest == vundefined()? to control tail policy. Is it acceptable to use different way to control the tail policy? Or we add the ta argument even if dest == vundefined()? is enough.

IMO, providing one way to control tail policy is more clear for users.

arcbbb:
> > I updated the document. I have two questions when modifying the document.
> > 
> > 2. Some intrinsics use `ta` argument and some intrinsics use `dest == vundefined()?` to control tail policy. Is it acceptable to use different way to control the tail policy? Or we add the `ta` argument even if `dest == vundefined()?` is enough.
> > 
> > It becomes
> > `ta` == true, we do not care about `dest`'s value.
> > `ta` == false, and `dest` == vundefined() => optimized to ta,ma.
> > `ta` == false, and `dest` != vundefined() => use tu,ma
> 
> If we have the dest operand and ta operand, will the compiler need to replace the dest with undef if agnostic is selected to give freedom to the register allocator and allow the computation to be dead code eliminated if the value passed to dest isn't used anywhere else?

I think maybe we could add an extra bit in `policy` to encode this compatibility,
like hardcoding the `policy` to be `TAIL_AGNOSTIC | COMPAT_V1` when forwarding `_m` intrinsics to `_mt` intrinsics.

Except for `policy & COMPAT_V1`, we have chances to replace dest with vundefined() for  `VTA=1`.

Hsiangkai:
In the latest update, I keep all the original intrinsics. Only add `_t` and `_mt` intrinsics to control tail policy.

Hsiangkai:
> > I updated the document. I have two questions when modifying the document.
> > 
> > 1. Do we need to control tail policy for
> > 
> > 
> > * Single-Width Floating-Point/Integer Type-Convert Instructions
> > * Widening Floating-Point/Integer Type-Convert Instructions
> > * Narrowing Floating-Point/Integer Type-Convert Instructions
> > 
> > If we keep the tail elements, does it mean parts of the elements are floating point and parts of the elements are integer? Does it make sense to keep tail elements for conversion intrinsics?
> 
> The tail elements would come from a different operand/register right? So they could be integer values if `vd` is a different register than `vs2`.
> 
> > 1. Some intrinsics use `ta` argument and some intrinsics use `dest == vundefined()?` to control tail policy. Is it acceptable to use different way to control the tail policy? Or we add the `ta` argument even if `dest == vundefined()?` is enough.
> > 
> > It becomes
> > `ta` == true, we do not care about `dest`'s value.
> > `ta` == false, and `dest` == vundefined() => optimized to ta,ma.
> > `ta` == false, and `dest` != vundefined() => use tu,ma
> 
> If we have the dest operand and ta operand, will the compiler need to replace the dest with undef if agnostic is selected to give freedom to the register allocator and allow the computation to be dead code eliminated if the value passed to dest isn't used anywhere else?

In the latest update, I use `_t` for intrinsics with `dest` to control tail elements. It may be less confusing according to the intrinsic name.

`vint8mf8_t vmv_v_v_i8mf8 (vint8mf8_t src, size_t vl);` is the original one.
`vint8mf8_t vmv_v_v_i8mf8_t (vint8mf8_t dest, vint8mf8_t src, size_t vl);` end with `_t` to let users know it is related to tail control and users could use `dest` as `vundefined()` to switch to `ta,ma`.

kito-cheng:
+1 for default `ta`, and add `_t` and `_mt` variant for `tu` and keep existing API, but I suppose user use `_t` or `_mt` means they want `tail undisturbed`, so we don't need extra option to control the policy?

Use vadd as example:
- TA: use vadd_vv_i32m1
- TU: use vadd_vv_i32m1_t or vadd_vv_i32m1_mt

arcbbb:
> +1 for default `ta`, and add `_t` and `_mt` variant for `tu` and keep existing API, but I suppose user use `_t` or `_mt` means they want `tail undisturbed`, so we don't need extra option to control the policy?
> 
> Use vadd as example:
> 
> * TA: use vadd_vv_i32m1
> * TU: use vadd_vv_i32m1_t or vadd_vv_i32m1_mt

IIUC, they look the same at IR level
for example, both TA & TU will be llvm.riscv.vadd.mask.nxv2i32.nxvvi32
```
declare <vscale x 2 x i32> @llvm.riscv.vadd.mask.nxv2i32.nxv2i32(
  <vscale x 2 x i32>,
  <vscale x 2 x i32>,
  <vscale x 2 x i32>,
  <vscale x 2 x i1>,
  i64, i64);
```

Hsiangkai:
> +1 for default `ta`, and add `_t` and `_mt` variant for `tu` and keep existing API, but I suppose user use `_t` or `_mt` means they want `tail undisturbed`, so we don't need extra option to control the policy?
> 
> Use vadd as example:
> 
> * TA: use vadd_vv_i32m1
> * TU: use vadd_vv_i32m1_t or vadd_vv_i32m1_mt

Initially, the introduced `ta` argument is used to avoid increasing the number of intrinsics. If we decide to add a new set of intrinsics for tail policy, it seems has no reason to keep the `ta` argument, does it? Unless we want to deprecate the `_m` intrinsics one day (it is not a fully defined interface).

Hsiangkai:
> > +1 for default `ta`, and add `_t` and `_mt` variant for `tu` and keep existing API, but I suppose user use `_t` or `_mt` means they want `tail undisturbed`, so we don't need extra option to control the policy?
> > Use vadd as example:
> > 
> > * TA: use vadd_vv_i32m1
> > * TU: use vadd_vv_i32m1_t or vadd_vv_i32m1_mt
> 
> IIUC, they look the same at IR level
> for example, both TA & TU will be llvm.riscv.vadd.mask.nxv2i32.nxvvi32
> 
> ```
> declare <vscale x 2 x i32> @llvm.riscv.vadd.mask.nxv2i32.nxv2i32(
>   <vscale x 2 x i32>,
>   <vscale x 2 x i32>,
>   <vscale x 2 x i32>,
>   <vscale x 2 x i1>,
>   i64, i64);
> ```

In LLVM IR level, we still need to add an additional argument for tail policy. It becomes

```
vadd_vv_i32m1_m -> @llvm.riscv.vadd.mask.nxv2i32.nxv2i32(..., vl, 1)
vadd_vv_i32m1_mt -> @llvm.riscv.vadd.mask.nxv2i32.nxv2i32(..., vl, 0)
```

kito-cheng:
> > > +1 for default `ta`, and add `_t` and `_mt` variant for `tu` and keep existing API, but I suppose user use `_t` or `_mt` means they want `tail undisturbed`, so we don't need extra option to control the policy?
> > > Use vadd as example:
> > > 
> > > * TA: use vadd_vv_i32m1
> > > * TU: use vadd_vv_i32m1_t or vadd_vv_i32m1_mt
> > 
> > 
> > IIUC, they look the same at IR level
> > for example, both TA & TU will be llvm.riscv.vadd.mask.nxv2i32.nxvvi32
> > ```
> > declare <vscale x 2 x i32> @llvm.riscv.vadd.mask.nxv2i32.nxv2i32(
> >   <vscale x 2 x i32>,
> >   <vscale x 2 x i32>,
> >   <vscale x 2 x i32>,
> >   <vscale x 2 x i1>,
> >   i64, i64);
> > ```
> 
> In LLVM IR level, we still need to add an additional argument for tail policy. It becomes
> 
> ```
> vadd_vv_i32m1_m -> @llvm.riscv.vadd.mask.nxv2i32.nxv2i32(..., vl, 1)
> vadd_vv_i32m1_mt -> @llvm.riscv.vadd.mask.nxv2i32.nxv2i32(..., vl, 0)
> ```

Yeah, and I believe that's kind of implementation detail we can hide, vsetvli is a good example, we have 3 argument for LLVM IR intrinsic, but we have only one option for C-level intrinsic:

```
#define vsetvl_e8mf8(avl) __builtin_rvv_vsetvli((size_t)(avl), 0, 5)
```

zakk0610:
Could you please update rvv-intrinsic-rfc.md to describe more about tail agnostic/undisturbed API?

LGTM for latest update.

Hsiangkai:
> Could you please update rvv-intrinsic-rfc.md to describe more about tail agnostic/undisturbed API?
> 
> LGTM for latest update.

Thanks. I have updated rvv-intrinsic-rfc.md.

zakk0610:
> Here is the proof-of-concept patch: https://reviews.llvm.org/D105092.
> 
> If all agree to define `_mt` intrinsics for masked operations with tail policy and keep `_m` intrinsics with tail agnostic as default tail policy, I could update the document accordingly.
> 
> In this PoC patch, I didn't model `maskedoff == vundefined()` behavior.

I think we also need to mention that the default tail policy for `_m` intrinsics is tail agnostic?

Hsiangkai:
> > Here is the proof-of-concept patch: https://reviews.llvm.org/D105092.
> > If all agree to define `_mt` intrinsics for masked operations with tail policy and keep `_m` intrinsics with tail agnostic as default tail policy, I could update the document accordingly.
> > In this PoC patch, I didn't model `maskedoff == vundefined()` behavior.
> 
> I think we also need to mention that the default tail policy for `_m` intrinsics is tail agnostic?

Added.

kito-cheng (approved):
LGTM

PR #103, opened 2021-07-02
Update overloading name for all miscellaneous functions.

zakk0610:
The one thing should be noted is in the same category function `vget` and
`vset` have different overloading naming rule, because `vget` need to encode
the return type in function name for distinct the different operation.


topperc (commented):
LGTM

PR #109, opened 2021-09-01
Add v1.0-rc1 examples

eopXD:
This PR wants to add some examples for rvv-spec v1.0-rc1.

Examples added:
- C intrinsic codes base on assembly examples in `riscv/riscv-v-spec`
- Other usage examples that demonstrate vector instructions. Usages like...
    - Element index instruction `vid` (rvv_index.c)
    - Reduction instruction `vfredsum` (rvv_reduce.c)
    - Masked arithmetic instruction (rvv_branch.c)
- A plain matmul with vector instructions


examples/rvv_strncpy.c
```diff
@@ -0,0 +1,62 @@
...
+
+  while (firstSetBit < 0 && count > 0) {
+    size_t vl = vsetvl_e8m8(count);
+    vl = min(n, vl);
+
+    size_t tmp;
+    vint8m8_t vec_src = vle8ff_v_i8m8(src, &tmp, vlmax);
+
+    vbool1_t stringTerminate = vmseq_vx_i8m8_b1(vec_src, 0, vlmax);
+    vbool1_t mask = vmsbf_m_b1(stringTerminate, vlmax);
```
zakk0610, line 20: use vmsif? "Set mask up to and including zero byte."

examples/rvv_strncpy.c
```diff
@@ -0,0 +1,62 @@
+#include "common.h"
+#include <riscv_vector.h>
+#include <string.h>
+
+size_t min(size_t a, size_t b) { return a < b ? a : b; }
+char *strncpy_vec(char *dst, char *src, size_t count) {
+  size_t vlmax = vsetvlmax_e8m8();
```
zakk0610, line 7: I think maybe having same logic implementation to spec assembly example would be better to see what's different between assembly and intrinsic.

examples/rvv_strncpy.c
```diff
@@ -0,0 +1,62 @@
...
+int main() {
+  const int N = 1320;
+  const uint32_t seed = 0xdeadbeef;
+  srand(seed);
+
+  // data gen
+  char s0[N];
+  char s1[] = "the quick brown fox jumps over the lazy dog";
+  gen_string(s0, N);
+  size_t count = strlen(s1) + rand() % 500;
```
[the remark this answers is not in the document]
eopXD: By the definition of strncpy in [cppreference](https://en.cppreference.com/w/cpp/string/byte/strncpy),
> If, after copying the terminating null character from src, count is not reached, additional null characters are written to dest until the total of count characters have been written.

Setting count larger than source length can test all codes written for strncpy.

examples/rvv_strncpy.c
```diff
@@ -0,0 +1,66 @@
...
+    vse8_v_i8m1_m(mask, dst, vec_src, new_vl);
+
+    count -= new_vl;
+    src += new_vl;
+    dst += new_vl;
+
+    first_set_bit = vfirst_m_b8(string_terminate, new_vl);
+  }
+
+  if (first_set_bit == -1)
```
eopXD, line 29: This is an extra branching instruction compared to the assembly implementation. Unlike assembly we can't insert a return if the two branching instructions don't perform a jump. (We also prefer not to use `goto`)

So an extra instruction is needed here. The extra logic "first_set_bit == -1` implies that the previous loop exited because count is reduced to zero.

eopXD: Thanks to @zakk, adjusted code and the extra logic is not needed.

zakk0610 (approved):
LGTM. Thanks!!

eopXD:
Thank you for reviewing this PR! Really appreciate it.

PR #114, opened 2021-09-23
Remove types that are not supported for ELEN=32

topperc:
Per the spec "For standard vector extensions with ELEN=32, fractional LMULs of 1/2 and 1/4 must be supported." and "For a given supported fractional LMUL setting, implementations must support SEW settings between SEWMIN and LMUL * ELEN, inclusive."


zakk0610 (approved):
LGTM, thanks.

PR #116, opened 2021-09-29
Add _t intrinsics for unmasked intrinsics with tail policy.

Hsiangkai:
Preliminary work for unmasked intrinsics with tail policy.

As pointed out in https://github.com/sifive/riscv-llvm-internal/issues/10, to use masked intrinsics with all-one mask will get worse code generated. If we have unmasked intrinsics with tail policy, users are able to write better code.


nick-knight (approved):
I am happy to finally have the ability to control tail policy --- thank you! 

My main use-case for this is to guarantee tail-undisturbed behavior in certain cases. Presently I do not have a use-case for guaranteeing tail-agnostic behavior, besides possible microarchitecture-specific performance optimizations. As such, I see your proposal as more general than necessary, and I'm motivated to share a counterproposal, in case others have similar thoughts.

Unmasked case: (`vop` is a placeholder for existing intrinsics like `vle16_v_u16m8`)
- `vop`: compiler is free to choose tail and mask policies;
- `vop_tu`: user wants tail-undisturbed; compiler is free to choose mask policy.

Masked case:
- `vop`: compiler is free to choose tail and mask policies;
- `vop_mu`: user wants mask-undisturbed; compiler is free to choose tail policy;
- `vop_tu`: user wants tail-undisturbed; compiler is free to choose mask policy;
- `vop_tumu`: user wants mask- and tail-undisturbed.

Note that "compiler is free to choose {tail,mask} policy" is _not_ the same as "user wants {tail,mask}-agnostic policy". It just means the user is "agnostic" to the choice the compiler makes. For example, on some microarchitectures we expect to see an additional performance cost to undisturbed behavior, so the compiler should prefer agnostic behavior if given the choice. On the other hand, if the user requests undisturbed behavior in certain places, then the compiler might be able to avoid `vset{i}vli` instructions by using undisturbed behavior elsewhere, even if the user didn't request it. These are decisions the programmer is entrusting to the compiler (and its cost model).

In summary, the major difference is that my counterproposal does not allow the user to guarantee {mask,tail}-agnostic policies. Your proposal --- which addresses only unmasked intrinsics --- is more general in this respect. (A minor difference is that yours is more programmer-friendly for dynamic tail policy, although I have yet to see anyone use that feature outside of context-switch code.) I haven't seen your proposal for the masked case, but I suspect it will end up being more complicated with both mask and tail policy.

intrinsic_funcs.md
```diff
@@ -197,6 +197,65 @@ vfloat64m1_t vle64_v_f64m1 (const float64_t *base, size_t vl);
 vfloat64m2_t vle64_v_f64m2 (const float64_t *base, size_t vl);
 vfloat64m4_t vle64_v_f64m4 (const float64_t *base, size_t vl);
 vfloat64m8_t vle64_v_f64m8 (const float64_t *base, size_t vl);
+vint8mf8_t vle8_v_i8mf8_t (vint8mf8_t dest, const int8_t *base, size_t vl, uint8_t ta);
```
topperc, line 200: The implementation in clang for the _mt intrinsics is using size_t not uint8_t for policy if I'm reading RISCVVEmitter.cpp correctly.

topperc:
(A minor difference is that yours is more programmer-friendly for dynamic tail policy, although I have yet to see anyone use dynamic `vtype` outside of context-switch code.) 

I don't think this proposal was intended to allow it to be dynamic. I believe the compiler will require it to be a constant.

rofirrim:
> My main use-case for this is to guarantee tail-undisturbed behavior in certain cases. Presently I do not have a use-case for guaranteeing tail-agnostic behavior, besides possible microarchitecture-specific performance optimizations.

> In summary, the major difference is that my counterproposal does not allow the user to guarantee the agnostic policies, just the undisturbed ones.

I don't think tail agnostic / mask agnostic can be guaranteed in any way given that the current wording of the spec is "may be overwritten with 1s, or with the result of the instruction in the case of mask-producing instructions except for mask loads"  (this is for tail agnostic but a similar text is used in mask agnostic case). So a hardware implementation is already compliant if it only implements tail undisturbed and completely ignores tail agnostic or mask agnostic.

In that sense this causes a bit of an asymmetry in the architecture: the undisturbed policy must be fulfilled but the agnostic policy is a hint.

This leads to your idea that we might be able to go away with a version of unmasked that doesn't enforce any specific tail policy. I think this what you suggest for the `vop` unmasked versions. I agree a compiler could exploit this extra bit of freedom as you mention.

Using your notation: for unmasked now we already have (unmasked) `vop` and (masked) `vop_mu`. From what I understand, Kai's proposal adds (unmasked) `vop_tx` (where a parameter determines whether `tx` is `ta` or `tu`). And then a later change could add (masked) `vop_txmx` (with two extra parameters determining the policies).

So, with Kai's proposal we end with (unmasked) `vop`, `vop_tx` and (masked) `vop_mu`, `vop_txmx`. If I got your proposal right we'd end with (unmasked) `vop`, `vop_tu` and (masked) `vop`, `vop_mu`, `vop_tu`, `vop_tumu`. It looks to me that a  proposal where the tail policy is controlled via a parameter provides a more moderated increase to the intrinsic API surface. Speciall given that anything including an undisturbed policy will have an extra `dest`/`maskedoff` so (masked) `vop_mu`, `vop_tu`, `vop_tumu` are identical interfaces which to me suggests using policy parameters (as in `vop_txmx`) might be a better choice.

Alternatively we could have (unmasked) `vop` and `vop_tu` and (masked) `vop_mu`, `vop_txmx` but some orthogonality gets lost between masked and unmasked under this scheme.

rofirrim:
> I don't think this proposal was intended to allow it to be dynamic. I believe the compiler will require it to be a constant.

That was my impression too. I think the use of parameters is just to avoid an explosion of an already very large API set.

nick-knight:
@rofirrim: Thanks for sharing your thoughts. The communication challenge here is there are two notions of agnosticism: the architectural notion (`ta`/`ma` policies, as defined in the spec), and that of the C programmer not caring whether the compiler generates code with `ta`/`ma` vs. `tu`/`mu`. Because of this challenge I'm not sure I'm communicating well.

Perhaps my biggest confusion is the behavior of @Hsiangkai's intrinsics when different `uint8_t` values are passed in `ta`. I didn't see this written down explicitly, I just guessed that it was meant to closely follow the behavior of the `vta` field in `vtype`, i.e., `ta == 0` implies `vta = 0` implies tail-undisturbed (a `vset{i}vli` with `tu`) and `ta != 0` implies `vta = 1` implies tail-agnostic (a `vset{i}vli` with `ta`). Based on this assumption, I then became concerned about a future hypothetical extension of Kai's proposal to masked instructions. Specifically, how would it express the case of a masked instruction that the programmer requires be tail-undisturbed (`tu`) but wants the compiler to decide the mask policy (`mu` or `ma`)? If I understand correctly, (what you call) `vop_txmx` would enable generating code with both `tu,ma` and `tu,mu`, but the programmer would have to decide. (The existing (what you call) `vop_mu` intrinsics also do not address this case.)

Alternatively --- and perhaps this is the failure in my understanding or communication --- @Hsiangkai 's intention is to allow the compiler to reinterpret the intrinsic arguments `ta != 0` and `ma != 0` in the "second notion" of agnosticism, meaning the user might pass `ta == 1` but the compiler generates code with `tu`. If this is the case, then I feel the API is somewhat redundant, with unmasked `vop` and `vop_t (`ta != 0`) being equivalent, and (I believe) other redundancies in the masked case.

zakk0610:
> Specifically, how would it express the case of a masked instruction that the programmer requires be tail-undisturbed (tu) but wants the compiler to decide the mask policy (mu or ma)?

For example:
`vint8m1_t vadd_vv_i8m1_mt (vbool8_t mask, vint8m1_t maskedoff, vint8m1_t op1, vint8m1_t op2, size_t vl, uint8_t ta);`
It could depend on maskedoff operand, if you are passing a `vundefined_* ()` and then compiler will choose `ma` for you, otherwise it choose `mu`.

maybe we need to update rfc document to make unmaksed intrinsic design become clear? 
BTW, sifive/riscv-llvm-internal#10 is private repo,  some people can not access it.

nick-knight:
@Hsiangkai I realize that I had never read 
https://github.com/riscv-non-isa/rvv-intrinsic-doc/blob/master/rvv-intrinsic-rfc.md#tail-policy-in-the-intrinsics
I guess this is not up-to-date with your current proposal? It would be good to include that update in this PR.

Hsiangkai:
> I am happy to finally have the ability to control tail policy --- thank you!
> 
> My main use-case for this is to guarantee tail-undisturbed behavior in certain cases. Presently I do not have a use-case for guaranteeing tail-agnostic behavior, besides possible microarchitecture-specific performance optimizations. As such, I see your proposal as more general than necessary, and I'm motivated to share a counterproposal, in case others have similar thoughts.
> 
> Unmasked case: (`vop` is a placeholder for existing intrinsics like `vle16_v_u16m8`)
> 
> * `vop`: compiler is free to choose tail and (irrelevant) mask policies;
> * `vop_tu`: user wants tail-undisturbed; compiler is free to choose (irrelevant) mask policy.
> 
> Masked case:
> 
> * `vop`: compiler is free to choose tail and mask policies;
> * `vop_mu`: user wants mask-undisturbed; compiler is free to choose tail policy;
> * `vop_tu`: user wants tail-undisturbed; compiler is free to choose mask policy;
> * `vop_tumu`: user wants mask- and tail-undisturbed.
> 
> Note that "compiler is free to choose {tail,mask} policy" is _not_ the same as "user wants {tail,mask}-agnostic policy". It just means the user is "agnostic" to the choice the compiler makes. For example, on some microarchitectures we expect to see an additional performance cost to undisturbed behavior, so the compiler should prefer agnostic behavior if given the choice. On the other hand, if the user requests undisturbed behavior in certain places, then the compiler might be able to avoid `vset{i}vli` instructions by using undisturbed behavior elsewhere, even if the user didn't request it. These are decisions the programmer is entrusting to the compiler (and its cost model).
> 
> In summary, the major difference is that my counterproposal does not allow the user to guarantee the agnostic policies, just the undisturbed ones. Your proposal --- which addresses only unmasked intrinsics --- is more general in this respect. (A minor difference is that yours is more programmer-friendly for dynamic tail policy, although I have yet to see anyone use dynamic `vtype` outside of context-switch code.) Also, I haven't seen how your proposal will generalize to the masked case, but I suspect it will end up being more complicated with both mask and tail policies. (For example, how would you handle what I called `vop_tu` above?)

Masked intrinsics already have tail policy argument in https://github.com/riscv-non-isa/rvv-intrinsic-doc/pull/101. That is why I only consider unmasked cases in the proposal.

Masked intrinsics have `_mt` version intrinsics with an additional tail policy argument. Users could use the argument to control tail agnostic/undisturbed. In the current design, we use maskedoff == vundefined()? to specify mask agnostic/undisturbed. However, there are some ta,tu/ma,mu combinations we could not express in the current design as described in https://github.com/riscv-non-isa/rvv-intrinsic-doc/blob/master/rvv-intrinsic-rfc.md#tail-policy. Especially for multiply-add intrinsics, dest could not be vundefined() as it is a part of the calculation inputs. So, I think we may need to extend the `ta` argument as `policy` argument and let users specify mask agnostic/undisturbed directly, not using maskedoff == vundefined? as an indicator. It is another issue.

Back to the proposal, I am also not sure if we need the additional `ta` argument in the unmasked cases. As @nick-knight said, it is redundant. We only need tail undisturbed behavior in these intrinsics. We could change the proposal to

```
vint8mf8_t vle8_v_i8mf8_t (vint8mf8_t dest, const int8_t *base, size_t vl);
```

When users use `_t` intrinsics, it is must be tail undisturbed.

Hsiangkai:
> > Specifically, how would it express the case of a masked instruction that the programmer requires be tail-undisturbed (tu) but wants the compiler to decide the mask policy (mu or ma)?
> 
> For example: `vint8m1_t vadd_vv_i8m1_mt (vbool8_t mask, vint8m1_t maskedoff, vint8m1_t op1, vint8m1_t op2, size_t vl, uint8_t ta);` It could depend on maskedoff operand, if you are passing a `vundefined_* ()` and then compiler will choose `ma` for you, otherwise it choose `mu`.
> 
> maybe we need to update rfc document to make unmaksed intrinsic design become clear? BTW, [sifive/riscv-llvm-internal#10](https://github.com/sifive/riscv-llvm-internal/issues/10) is private repo, some people can not access it.

Thanks. I didn't notice the issue [sifive/riscv-llvm-internal#10](https://github.com/sifive/riscv-llvm-internal/issues/10) comes from private repo. Could you help to post it in this repo? Thanks.

Hsiangkai:
> > My main use-case for this is to guarantee tail-undisturbed behavior in certain cases. Presently I do not have a use-case for guaranteeing tail-agnostic behavior, besides possible microarchitecture-specific performance optimizations.
> 
> > In summary, the major difference is that my counterproposal does not allow the user to guarantee the agnostic policies, just the undisturbed ones.
> 
> I don't think tail agnostic / mask agnostic can be guaranteed in any way given that the current wording of the spec is "may be overwritten with 1s, or with the result of the instruction in the case of mask-producing instructions except for mask loads" (this is for tail agnostic but a similar text is used in mask agnostic case). So a hardware implementation is already compliant if it only implements tail undisturbed and completely ignores tail agnostic or mask agnostic.
> 
> In that sense this causes a bit of an asymmetry in the architecture: the undisturbed policy must be fulfilled but the agnostic policy is a hint.
> 
> This leads to your idea that we might be able to go away with a version of unmasked that doesn't enforce any specific tail policy. I think this what you suggest for the `vop` unmasked versions. I agree a compiler could exploit this extra bit of freedom as you mention.
> 
> Using your notation: for unmasked now we already have (unmasked) `vop` and (masked) `vop_mu`. From what I understand, Kai's proposal adds (unmasked) `vop_tx` (where a parameter determines whether `tx` is `ta` or `tu`). And then a later change could add (masked) `vop_txmx` (with two extra parameters determining the policies).
> 
> So, with Kai's proposal we end with (unmasked) `vop`, `vop_tx` and (masked) `vop_mu`, `vop_txmx`. If I got your proposal right we'd end with (unmasked) `vop`, `vop_tu` and (masked) `vop`, `vop_mu`, `vop_tu`, `vop_tumu`. It looks to me that a proposal where the tail policy is controlled via a parameter provides a more moderated increase to the intrinsic API surface. Speciall given that anything including an undisturbed policy will have an extra `dest`/`maskedoff` so (masked) `vop_mu`, `vop_tu`, `vop_tumu` are identical interfaces which to me suggests using policy parameters (as in `vop_txmx`) might be a better choice.
> 
> Alternatively we could have (unmasked) `vop` and `vop_tu` and (masked) `vop_mu`, `vop_txmx` but some orthogonality gets lost between masked and unmasked under this scheme.

In the current `_mt` intrinsics, it should be able to cover (masked) `vop_mu`, `vop_tu`, `vop_tumu`.

intrinsic_funcs.md
Hsiangkai: You are right. I need to update it to size_t.

Hsiangkai:
> @Hsiangkai I realize that I had never read https://github.com/riscv-non-isa/rvv-intrinsic-doc/blob/master/rvv-intrinsic-rfc.md#tail-policy-in-the-intrinsics I guess this is not up-to-date with your current proposal? It would be good to include that update in this PR.

I agree with you. I will update the RFC document to include unmasked cases in this PR.

zakk0610:
> Thanks. I didn't notice the issue sifive/riscv-llvm-internal#10 comes from private repo. Could you help to post it in this repo? Thanks.

https://godbolt.org/z/ndcTfh87h is a example showing  "_mt  intrinsics with all-one mask will get worse code generated."
```
#include <riscv_vector.h>
vint8m1_t int8_load(const int8_t *base, size_t vl) {
   size_t vlmax = vsetvlmax_e8m1();
   vint8m1_t vzero = vmv_v_x_i8m1(0, vlmax);
   vbool8_t mask = vmset_m_b8(vl);
   return vle8_v_i8m1_mt(mask, vzero, base, vl, VE_TAIL_UNDISTURBED);
 }
 ```
 
 ```
 int8_load:           
        vsetvli a2, zero, e8, m1, ta, mu
        vmv.v.i v8, 0
        vsetvli zero, a1, e8, m1, ta, mu
        vmset.m v0
        vsetvli zero, zero, e8, m1, tu, mu
        vle8.v  v8, (a0), v0.t
        ret
```
ideally compiler could generate optimized code like below:
```
 int8_load:                                        
         vsetvli a2, zero, e8,m1,ta,mu                        
         vmv.v.i v8, 0                                                   
         vsetvli a1, a1, e8,m1,tu,mu              
         vle8.v  v8, (a0)        
         ret          
```

rofirrim:
> In the current `_mt` intrinsics, it should be able to cover (masked) `vop_mu`, `vop_tu`, `vop_tumu`.
> I agree with you. I will update the RFC document to include unmasked cases in this PR.

Thanks @Hsiangkai now I understand the reason for this change. We add a new case `_t` (looks like we can't leverage on the "all-ones" case for `_mt`). 

On the bright side, this approach is a bit more predictable as we don't rely on the compiler emitting an unmasked operation for us in a special case. The downside is a new set of intrinsics `_t` but I understand we're OK with this.

nick-knight (approved):

PR #122, opened 2021-11-02
simply example with vfmul.vv instead of vfmacc.vv

lhtin:
From the mean of the example, I think the no need to use `vfmacc`(https://github.com/riscv-non-isa/rvv-intrinsic-doc/blob/master/examples/rvv_matmul.c#L30). Because the below call `vfredsum_vs_f64m1_f64m1` will sum the multiple results.

### DIFF
--- a/examples/rvv_matmul.c
+++ b/examples/rvv_matmul.c
@@ -27,7 +27,7 @@ void matmul(double **a, double **b, double **c, int n, int m, int o) {
         vfloat64m1_t vec_a = vle64_v_f64m1(ptr_a, vl);
         vfloat64m1_t vec_b = vle64_v_f64m1(ptr_b, vl);
 
-        vec_s = vfmacc_vv_f64m1(vec_s, vec_a, vec_b, vl);
+        vec_s = vfmul_vv_f64m1(vec_s, vec_a, vec_b, vl);
       }
 
       vfloat64m1_t vec_sum;


nick-knight:
I think this PR introduces a bug, except in the special case O <= VLEN / 64.

examples/rvv_matmul.c
```diff
@@ -27,7 +27,7 @@ void matmul(double **a, double **b, double **c, int n, int m, int o) {
         vfloat64m1_t vec_a = vle64_v_f64m1(ptr_a, vl);
         vfloat64m1_t vec_b = vle64_v_f64m1(ptr_b, vl);
 
-        vec_s = vfmacc_vv_f64m1(vec_s, vec_a, vec_b, vl);
+        vec_s = vfmul_vv_f64m1(vec_s, vec_a, vec_b, vl);
```
topperc, line 30: vfmul_vv_f64m1 only takes 3 operands not 4. You need vfmacc to accmulate into vec_s across multiple loop iterations.

eopXD:
`vfmacc_vv_f64m1` is used to accumulate multiplications into `vec_s` and the `vfredsum_vs_f64m1_f64m1` is used to reduce the multiplications of slots in `vec_s` into single slot in `vec_sum` then convert it into a double.

You can check out the rvv-spec [here](https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc).

examples/rvv_matmul.c
lhtin: thanks for your review and specify the bug. I'm very sorry for this mistake PR and I will close the PR. sorry for this again.

lhtin:
> `vfmacc_vv_f64m1` is used to accumulate multiplications into `vec_s` and the `vfredsum_vs_f64m1_f64m1` is used to reduce the multiplications of slots in `vec_s` into single slot into `vec_sum` then convert it to a double.
> 
> You can check out the rvv-spec [here](https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc).

I get your idea and am sorry for this wrong PR. I will close it.

PR #123, opened 2021-11-18
fix the order of mask and maskedoff from the tail-policy table

lhtin:
The order of mask and maskedoff args in [the tail-policy table](https://github.com/riscv-non-isa/rvv-intrinsic-doc/blob/master/rvv-intrinsic-rfc.md#tail-policy-in-the-intrinsics) is different from the prototype of the intrinsic function. So need be consistent.

old:
```
vadd_vv_<ty>_mt(maskedoff, mask, vs2, vs1, vl, VE_TAIL_AGNOSTIC)
```

change to:
```
vadd_vv_<ty>_mt(mask, maskedoff, vs2, vs1, vl, VE_TAIL_AGNOSTIC)
```


Hsiangkai (approved):
LGTM.

PR #128, opened 2021-12-20
Fix typos in Single-Width Integer Multiply-Add.

kuanlinchentw:
Fixing typos in 12.13. Vector Single-Width Integer Multiply-Add Operations.


eopXD:
Looks good, thank you!

PR #130, opened 2022-02-06
Define the availability of the intrinsics (via riscv_vector.h)

cmuellner:
The example code indicates that the header file <riscv_vector.h> needs
to be included. Let's document this behaviour and mention that
intrinsics require their corresponding instruction to be available.


eopXD (commented):
Looks good, I think this can be merged.

kito-cheng (approved):
Thanks!

Issue #131, opened 2022-02-07
Question about the LLVM emits vsetvl with `tu, mu` when compiling the `vmacc_vv_i8m1` intrinsics


lhtin:
Hi all,

I have a question about that the LLVM compiler emits vsetvli with `tu, mu`  when compiling the `vmacc_vv_i8m1` intrinsics (i.e. has dest parameter). Because from the RFC doc, it should emit `ta, mu`. And there should be a corresponding `_tu` intrinsic which should emit `tu, mu`. And from my experience about using intrinsic, it can reduce the number of the vsetvli instruction.

[Here](https://godbolt.org/z/c4zoKvcje) can recurrent.

C code:
```c
#include <riscv_vector.h>

vint8m1_t vmacc_test(vint8m1_t dest, vint8m1_t op1, vint8m1_t op2, size_t vl)
{
    vint8m1_t t = vadd_vv_i8m1(op1, op2, vl);
    return vmacc_vv_i8m1(dest, t, op2, vl);
}
```

RVV Assembly:
```
vmacc_test:
        vsetvli zero, a0, e8, m1, ta, mu
        vadd.vv v9, v9, v10
        vsetvli zero, zero, e8, m1, tu, mu
        vmacc.vv        v8, v9, v10
        ret
```

Best regards,
Tin

zakk0610:
upstream LLVM had [reverted ](https://reviews.llvm.org/rG7ccd31c900edecb1ec3eb75bb6f917dbe674bc72)"_tu" and "_mt" intrinsics supported due to adding those new set of intrinsics increase clang binary size/clang compilation time. 
https://reviews.llvm.org/D110684#3060191
so before LLVM support that, the default behavior for vamcc are still always tail undisturbed.

BTW, we want to propose this new design https://github.com/riscv-non-isa/rvv-intrinsic-doc/issues/120 for tail/mask policy and it's still under reviewing and developing on phabricator. (Ex.  https://reviews.llvm.org/D111617, [D117681](https://reviews.llvm.org/D117681)  and /[D117989](https://reviews.llvm.org/D117989)).

fabianpedd:
[...]

Issue #120, opened 2021-10-07
Propose a new set of intrinsics considering tail and mask agnostic/undisturbed policy.


Hsiangkai:
In PR #101, we propose `_mt` intrinsics to introduce tail policy argument without adding too many intrinsics. However, there are several problems in the design.

* There is no intrinsic for tu,ma combination.
* There is no intrinsic for ta,ma for multiply-add intrinsics because the destination of multiply-add could not be vundefined().
* We use all-one mask and vundefined() to control the tail/mask policy. It is not an obvious way for users.
* The code gen quality is not good for unmasked tail undisturbed intrinsics.

For the last point, we propose PR #116 to introduce `_tu` variant for unmasked intrinsics.

There are two styles in the current intrinsics. For unmasked ones, we append the suffix `_tu` to control the policy. For masked ones, we use the additional `ta` argument plus some implication rules to control the policy.

Ideally, we should make the design consistent. So, we are thinking about to append {tu/ta}, {mu/ma} to the intrinsic names. There are some reasons to do it.

* {tu/ta}, {mu/ma} are parts of vtype. We already encode part of the vtype, e.g., i8m1, into the intrinsic names. It makes sense to encode {tu/ta}, {mu/ma} in the intrinsic names.
* It is apparent to users to use it. There is no implication rules for all-one mask and maskedoff == vundefined().
* It makes the design consistent between masked and unmasked intrinsics.

The only problem of the proposal is it will bring a whole bunch set of intrinsics. However, if we could reduce the header size of intrinsics. It should not be a serious problem. From https://reviews.llvm.org/D103228. It should be a solvable problem.

So, we propose the following new set of intrinsics for tail/mask policy.

Masked? | TU? | MU? | Intrinsic
---|---|---|---
No | No | N/A | `vadd_vv_i8m1_ta(vs2, vs1, vl)` and `vadd_vv_i8m1(vs2, vs1, vl)`<sup>1</sup>
No | Yes | N/A | `vadd_vv_i8m1_tu(merge,vs2, vs1, vl)`
Yes | No | No | `vadd_vv_i8m1_tama(mask, vs2, vs1, vl)`
Yes | No | Yes | `vadd_vv_i8m1_tamu(mask, merge, vs2, vs1, vl)`
Yes | Yes | No | `vadd_vv_i8m1_tuma(mask, merge, vs2, vs1, vl)`
[...]

PR #133, opened 2022-02-08
Remove _mt since that is obsolete, and gonna removed soon.

kito-cheng:
We'll have new API for that.

https://github.com/riscv-non-isa/rvv-intrinsic-doc/issues/131


zhongjuzhe:
> 

Can you update the intrinsics with new api? Are the new api changed to as https://github.com/riscv-non-isa/rvv-intrinsic-doc/issues/120 said. I want to make my GCC implementation consistent with the latest api documents. Thanks~

kito-cheng:
Yeah, that would be next step :)

kito-cheng:
Changes:
- Update `rvv-intrinsic-rfc.md`

Issue #126, opened 2021-11-23
Function prototype of vmerge and vadc-family instructions


eopXD:
I found that although the two operations share the same suffix `vvm` / `vxm`, their function prototype are inconsistent unlike `vv` / `vx` operations. Is there a specific reason behind this?

```
vint32m1_t vadc_vvm_i32m1 (vint32m1_t op1, vint32m1_t op2, vbool32_t carryin, size_t vl);
vint16m1_t vmerge_vvm_i16m1 (vbool16_t mask, vint16m1_t op1, vint16m1_t op2, size_t vl);
```

zakk0610:
The intrinsic names are as close as the assembly mnemonics. 
`MNEMONIC ::= Instruction name in v-ext specification. Replace '.' with '_'.`
`vadc`/`vmerge` are `vadc.vvm`/`vmerge.vvm`.

eopXD:
Thanks for the reply.

lhtin:
> The intrinsic names are as close as the assembly mnemonics. `MNEMONIC ::= Instruction name in v-ext specification. Replace '.' with '_'.` `vadc`/`vmerge` are `vadc.vvm`/`vmerge.vvm`.

I have another question, why vmerge and vader's mask parameter positions different?

nick-knight:
I actually have the same question. I think this was the original question, and I don't think it was addressed.

The `v*{ad,sb}c.v*m` instructions have a similar property to the `vmerge.*` instructions: they are encoded as masked, but operate as if they were unmasked, and treat the input mask as another input vector. (This is why the assembly mnemonics use `v0` instead of `v0.t`.) I think it is inconsistent that the `vmerge_*` intrinsics have the `mask` parameter first, while the `v*{ad,sb}c_v*m_*` intrinsics instead have a `carryin`/`borrowin` parameter third. I think it would make more sense if `vmerge` had its `mask` argument third.

zakk0610:
@nick-knight @lhtin Thanks, it makes sense to me.
I'm trying to remember something from https://github.com/riscv-non-isa/rvv-intrinsic-doc/issues/29.
It seems we make an exception rule for `vmerge` prototype, let it looks like vmaskop ? va : vb which familiar to C. 
I'm ok to update argument order, and we may need to change `vcompress` too.

PR #104, opened 2021-07-02
Rename argument name for mac family operations.

zakk0610:
// result = (op1 * op2) + acc
vfloat16mf4_t vfmacc_vv_f16mf4 (vfloat16mf4_t acc, vfloat16mf4_t op1,
vfloat16mf4_t op2, size_t vl);

// result = (op1 * acc) + op2
vfloat16mf4_t vfmadd_vv_f16mf4 (vfloat16mf4_t acc, vfloat16mf4_t op1,
vfloat16mf4_t op2, size_t vl);

Above are old interfaces and the argument  name `acc` is not clear for vfmadd functions.
I think maybe align the argument name with assembly operand name is more appropriate.

PR #137, opened 2022-02-23
[RFC] Support a new set of policy intrinsic functions.

zakk0610:
Support the policy intrinsic functions to control tail
and mask policy.

Some operations would not take tail and mask value,
so they would not have policy intrinsic functions.
So in this proposal, I would like to keep the current policy
rule for all non-policy intrinsic functions (ex. vadd_vv_i8m1
and vadd_vv_i8m1_m). It is the only one different thing comparing
with proposal in https://github.com/riscv-non-isa/rvv-intrinsic-doc/issues/120

Please see [rvv-intrinsic-rfc.md ](https://github.com/riscv-non-isa/rvv-intrinsic-doc/blob/new-policy-proposal/rvv-intrinsic-rfc.md#policy-intrinsic-functions) for more detail.

There are some special cases list as below:

1. vadc/vsbc support `_ta` and _`tu`.
2. vmadc/vmsbc support `_ta`.
3. Vector Integer and Floating-Point Comparison Instructions support `_ta`, `_tamu` and `_tama`.
4. Vector Integer and Floating-Point Move support `_ta` and `_tu`.
5. vlm.v support `_ta`.
6. vmsbf.m/vmsif.m/vmsof.m support `_ta`, `_tama` and `_tamu`.
7. vmv.s.x/vfmv.s.f support `_ta` and `_tu`.
8. Vector Integer and Floating-Point Merge Instructions support `_ta` and `_tu`.
9. Vector Compress Instruction support `_ta` and `_tu`.
10. Vector Store, vfirst.m, vmv.x.s, vfmv.f.s, vcpop.m and Miscellaneous Functions don't have don't have policy intrinsic functions.
11. vslideup and vslidedown don't have dest operand in `_ta` and `_tama`.

For overloading functions, encode the policy in the end of function name.
For example, vadd would have `vadd_ta`, `vadd_tu`, `vadd_tama`, `vadd_tamu`, `vadd_tuma` and `vadd_tumu`.


zhongjuzhe:
Thanks for the new PR. For this proposal, I have a question:
For *_tu of segement load, for example: void vlseg2e8_tu (vint8mf8_t *v0, vint8mf8_t *v1, const int8_t *base, size_t vl);
   Why there is no "undisturbed" in these intrinsic. I see vint8mf8_t vle8_v_i8mf8_tu (vint8mf8_t undisturbed, const int8_t *base, size_t vl);   has a "undisturbed" argument.

lhtin:
In my opinions:

0. The core is to support the corresponding policy intrinsics only if the corresponding instructions need it.
1. Intrinsics whose corresponding instructions **ignore tail policy and ignore mask policy** should only support `normal` intrinsics. e.g. `vmadc_vvm` 
2. Intrinsics whose corresponding instructions **ignore tail policy and need mask policy** should support `normal`, `_m`, `_ma`, `_mu` intrinsics. e.g. `vmseq_vv`, `vmseq_vv_m`, `vmseq_vv_ma`, `vmseq_vv_mu`
3. Intrinsics whose corresponding instructions **need tail policy and ignore mask policy** should support `normal`, `_ta`, `_tu` intrinsics. e.g. `vmerge_vvm`, `vmerge_vvm_ta`, `vmerge_vvm_tu`
4. Intrinsics whose corresponding instructions **need tail policy and need mask policy** should support all intrinsics(i.e. `normal`, `_ta`, `_tu`, `_m`, `_tama`, `_tamu`, `_tuma`, `_tumu`). e.g. `vadd_vv`, `vadd_vv_ta`, `vadd_vv_tu`, `vadd_vv_m`, `vadd_vv_tama`, `vadd_vv_tamu`, `vadd_vv_tuma`, `vadd_vv_tumu`.
5. Compiler choose the suitable policy for the unspecified intrinsics. e.g. `_ma` intrinsics can have `ta` or `tu` tail policy. For backward compatibility, `_m` intrinsics can always use `tumu` policy.

Based on the above point of view, I have the following questions:

1. Why vmadc/vmsbc need support `_ta` intrinsics? I think vmadc/vmsbc without `_ta` is enough. Because it ignores mask and tail policy for vmadc/vmsbc from the rvv spec[1]. So let the compiler choose the most suitable one. vlm.v is likes it.
2. The support of `_tamu` and `_tama` for Vector Integer and Floating-Point Comparison Instructions has the same confusion like 1. And I think these instructions should support `_ma` and `_mu` intrinsics instead. vmsbf.m/vmsif.m/vmsof.m are like it.
3. Can you talk more about compatibility issues? For example, for the redsum intrinsic, what is the influence of removing the vd parameter from the `redsum_vv_i8m1` intrinsics? How widely is this intrinsics used today?


[1] https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc#vector-integer-add-with-carry-subtract-with-borrow-instructions

nick-knight:
@zhongjuzhe :
> For *_tu of segement load, for example: void vlseg2e8_tu (vint8mf8_t *v0, vint8mf8_t *v1, const int8_t *base, size_t vl);
Why there is no "undisturbed" in these intrinsic. I see vint8mf8_t vle8_v_i8mf8_tu (vint8mf8_t undisturbed, const int8_t *base, size_t vl); has a "undisturbed" argument.

I believe the vector register pair passed (by pointer) in `v0` and `v1` serves the role of `undisturbed`.

nick-knight:
I agree with @lhtin: there are several classes of instructions whose behavior is always tail-agnostic (regardless of `vtype.vta`), so there is really no need to have `_ta` in the intrinsics names. It is true that removing `ta` causes some name collisions with the legacy API, but I don't think any of these collisions cause incompatibilities.

nick-knight:
@lhtin : 
> For example, for the redsum intrinsic, what is the influence of removing the vd parameter from the redsum_vv_i8m1 intrinsics? How widely is this intrinsics used today?

You mentioned this before in https://github.com/riscv-non-isa/rvv-intrinsic-doc/issues/120.

I believe your proposal is to change the legacy API so that, for example, `vredsum_vs_i8m1(vd, vs1, vs2, vl)` becomes equivalent to the (new/proposed) intrinsic `vredsum_vs_i8m1_ta(vs1, vs2, vl)`, losing the `vd` argument. While I support this proposal, I think this conversation can be separated from https://github.com/riscv-non-isa/rvv-intrinsic-doc/pull/137 (this PR). In particular, your proposal concerns changing the "legacy" API, rather than extending the API like this PR does. Do you agree this discussion can be separated?

rvv-intrinsic-rfc.md
```diff
@@ -348,70 +348,83 @@ vmerge.vvm vd, vs2, vs1, v0:
...
+Yes | No | No | `vmacc_vv_i8m1_tama(mask, vd, vs1, vs2, vl)`
+Yes | No | Yes| `vmacc_vv_i8m1_tamu(mask, vd, vs1, vs2, vl)`
+Yes | Yes| No | `vmacc_vv_i8m1_tuma(mask, vd, vs1, vs2, vl)`
+Yes | Yes| Yes | `vmacc_vv_i8m1_tumu(mask, vd, vs1, vs2, vl)` and `vmacc_vv_i8m1_m(mask, vd, vs1, vs2, vl)`
+
+For reduction intrinsics which may only use tail value.
+
+Masked?  | TU? | MU? | Intrinsic
+--- |--- |--- |---
+No  | No | N/A| `redsum_vs_i8m1_ta(vs1, vs2, vl)`
```
nick-knight, line 405: Missing `v` in the intrinsics names in this table

rvv-intrinsic-rfc.md
```diff
@@ -348,70 +348,83 @@ vmerge.vvm vd, vs2, vs1, v0:
...
+No  | Ye | N/A| `redsum_vs_i8m1_tu(vd, vs1, vs2, vl)` and `redsum_vs_i8m1(vd, vs1, vs2, vl)`
```
nick-knight, line 406: `Ye` -> `Yes`

rvv-intrinsic-rfc.md
```diff
@@ -348,70 +348,83 @@ vmerge.vvm vd, vs2, vs1, v0:
...
+Yes | No | No | `redsum_vs_i8m1_tama(mask, vs1, vs2, vl)`
+Yes | No | Yes| `redsum_vv_i8m1_m(mask, vd, vs1, vs2, vl)` if vd is `vundefined()`
+Yes | Yes| No | `redsum_vs_i8m1_tuma(mask, vd, vs1, vs2, vl)`
+Yes | Yes| Yes| `redsum_vs_i8m1_m(mask, vd, vs1, vs2, vl)`
+
+There are some special cases list as below:
+
+1. vadc/vsbc support `_ta` and _`tu`.
+2. vmadc/vmsbc support `_ta`.
```
nick-knight, line 415: I think that special cases (2), (3), (5), and (6) could all omit `ta`, since they are always tail agnostic. It is true that some of these names collide with the legacy API, but I don't think any of these collisions cause incompatibilities.

rvv-intrinsic-rfc.md
```diff
@@ -348,70 +348,83 @@ vmerge.vvm vd, vs2, vs1, v0:
...
-| Yes     | No                   | Yes                       | No support.
-| Yes     | Yes                  | Yes                       | No support.
-| Yes     | Yes                  | No                        | No support. Tail undisturbed and maskedoff agnostic is likely rare.
+Masked?  | TU? | MU? | Intrinsic
+--- |--- |--- |---
+No  | No | N/A| `vadd_vv_i8m1_ta(vs2, vs1, vl)` and `vadd_vv_i8m1(vs2, vs1, vl)`
+No  | Yes| N/A| `vadd_vv_i8m1_tu(undisturbed,vs2, vs1, vl)`
+Yes | No | No | `vadd_vv_i8m1_tama(mask, vs2, vs1, vl)`
+Yes | No | Yes| `vadd_vv_i8m1_tamu(mask, merge, vs2, vs1, vl)` and `vadd_vv_i8m1_m(mask, merge, vs2, vs1, vl)` if merge is `vundefined()`
+Yes | Yes| No | `vadd_vv_i8m1_tuma(mask, merge, vs2, vs1, vl)`
```
nick-knight, line 382: I'm not clear why you call it `undisturbed` in the `tu` case, and `merge` in the `tuma` case. Suggest using same thing everywhere.

rvv-intrinsic-rfc.md
```diff
@@ -348,70 +348,83 @@ vmerge.vvm vd, vs2, vs1, v0:
 vint8m1_t vmerge_vvm_i8m1_m(vbool8_t mask, vint8m1_t vs2, vint8m1_t vs1, size_t vl);
```
nick-knight, line 348: As suggested in https://github.com/riscv-non-isa/rvv-intrinsic-doc/issues/126, I think we should consider moving the `mask` argument for consistency with other intrinsics. We could do this here in the new "policy" intrinsics, since they use `_ta` and `_tu`, and update the legacy API later. Just a proposal.

nick-knight (approved):

zhongjuzhe:
I have another question: I see vamo instrucions have been deleted in SPEC.  Why the RVV intrinsic doc still keep these intrinsics?

lhtin:
@nick-knight I agree to separate new discussion about the breaking change proposal. 

What about question 2? Is there a need for `_tama` and `_tamu` to compare intrinsics? From the rvv spec[1], compare instructions ignore the tail policy. So I think there is no need for `t[au]` in intrinsics name and `_ma` and `_mu` is enough.

[1] https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc#vector-integer-compare-instructions

zakk0610:
> I agree with @lhtin: there are several classes of instructions whose behavior is always tail-agnostic (regardless of `vtype.vta`), so there is really no need to have `_ta` in the intrinsics names. It is true that removing `ta` causes some name collisions with the 
legacy API, but I don't think any of these collisions cause incompatibilities.

IMPO, I feel "ignore the tail policy" not equal to "always operate under a tail-agnostic policy", so it's why I add `_ta` for those operations.  But I agree with your opinion, I will update RFC later.

> What about question 2? Is there a need for _tama and _tamu to compare intrinsics? From the rvv spec[1], compare instructions ignore the tail policy. So I think there is no need for t[au] in intrinsics name and _ma and _mu is enough.

IMPO, I would like to keep the `t[au]` to make less naming divergent, but I didn't have strong opinion here.

> I have another question: I see vamo instrucions have been deleted in SPEC. Why the RVV intrinsic doc still keep these intrinsics?

I will open another PR for removing `vamo` intrinsic functions.

@nick-knight @lhtin @zhongjuzhe thanks for reviewing, I will update PR later.

lhtin:
@zakk0610 

> IMPO, I feel "ignore the tail policy" not equal to "always operate under a tail-agnostic policy"

"ignore the tail policy" means the compiler can set any, it doesn't affect the behavior of the instruction. and
"always operate under a tail-agnostic policy" refers to the behavior of the hardware.

> IMPO, I would like to keep the t[au] to make less naming divergent, but I didn't have strong opinion here.

My main concern is that the compiler can make better decisions if the tail policy is not specified. If `_tama` is specified, the compiler can only set it to `ta`, otherwise, it does not meet the user's expectations. This leaves more optimization possibilities for the compiler.

nick-knight:
>  If _tama is specified, the compiler can only set it to ta, otherwise, it does not meet the user's expectations.

As a user, I have a different expectation, which I tried to explain here: https://github.com/riscv-non-isa/rvv-intrinsic-doc/issues/120#issuecomment-1041931451

lhtin:
> As a user, I have a different expectation, which I tried to explain here: [#120 (comment)](https://github.com/riscv-non-isa/rvv-intrinsic-doc/issues/120#issuecomment-1041931451)

I think ta is more unnecessary in intrinsics name if ta not mean the operand of vsetvl instruction. More ever, my point is the purpose of encoding t[au] and m[au] into the intrinsic name is to allow the user to precisely control the parameters of the corresponding vsetvl.

kito-cheng:
I am withdraw the proposal of changing legacy interface for reduction, I prefer don't break the interface if possible, and just let compiler to optimize if `vd` is `vundefined`.

kito-cheng:
> I think ta is more unnecessary in intrinsics name if ta not mean the operand of vsetvl instruction. More ever, my point is the purpose of encoding t[au] and m[au] into the intrinsic name is to allow the user to precisely control the parameters of the corresponding vsetvl.

+1 for removing `t[au]` for compare instructions.

lhtin:
@zakk0610 

> okay, if we decide to remove t[au] for compare instructions, I think we also need to remove m[au] for masked reduction instructions because they do not care the value in inactive elements. then we need to have a new naming rule for masked operation due to conflict naming with unmasked one. how about `m_tu` or `m_ta`?
> 
> Masked?	TU?	MU?	Intrinsic
> No	No	N/A	`vredsum_vs_i8m1_ta(vs1, vs2, vl)`
> No	Yes	N/A	`vredsum_vs_i8m1_tu(vd, vs1, vs2, vl)`
> Yes	No	No	`vredsum_vs_i8m1_m_ta(mask, vs1, vs2, vl)`
> Yes	No	Yes	`vredsum_vv_i8m1_m(mask, vd, vs1, vs2, vl)` if vd is `vundefined()`
> Yes	Yes	No	`vredsum_vs_i8m1_m_tu(mask, vd, vs1, vs2, vl)`
> Yes	Yes	Yes	`vredsum_vs_i8m1_m(mask, vd, vs1, vs2, vl)`

From my previous review comment, I think `vredsum_vs_i8m1_m_ta` and `vredsum_vs_i8m1_m_tu`'s `MU?` value should be `N/A` instead of `No`. That's to say let the compiler choose the best one.

> IMPO, compiler could translate TA/MA to TU/MU if there is any benefit no matter what users give., but it seems users would get surprised the changed because they are using intrinsic with specific policy. Is it? If yes, does it also mean compiler could not have combine optimization for intrinsic functions, right?

Yes, I think that's ok. More detail, from the discussion, there should be two ways at present:

1. **For intrinsics that do not specify a specific policy (N/A), the compiler decides which policy to use.** But for those specified in the intrinsic name (agnostic or undisturbed policy), the compiler must be set to be the same.
2. **For intrinsics that do not specify a specific (N/A) or an agnostic policy, the compiler decides which policy to use.** But for policies specified as undisturbed in the intrinsic name, the compiler must be set to be the same.

I prefer the first way. I think if the user chooses a policy, that's what he/she needs. Otherwise, he/she should choose the intrinsics that do not specify the policy.

lhtin:
@zakk0610 Another question would like to ask, why put the new intrinsics in the policy_funcs directory instead of merging them together?

rvv-intrinsic-rfc.md
```diff
@@ -348,70 +348,87 @@ vmerge.vvm vd, vs2, vs1, v0:
...
 ```
 
-## Tail Policy in the Intrinsics<a name="tail-policy"></a>
+## Policy Intrinsic Functions<a name="policy"></a>
 
-The naming rule of intrinsics with tail policy is
+Users could use policy intrinsic functions to control tail and inactive elements.
+Enocde the policy explicitly in function name suffix if operations support those policies.
```
eopXD, line 354: `Ecocde` -> `Encode`.

eopXD:
Replying to @lhtin 's comment on following the agnostic policy:

According to v-spec (under section [3.4.3](https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc#343-vector-tail-agnostic-and-vector-mask-agnostic-vta-and-vma)), the agnostic behavior is implementation defined to either retain the original value (which is the undisturbed behavior) or set to all 1s.

In my opinion the user should not depend on the behaviors if specified agnostic (and I think the 2nd note under the section mentioned supports this).

Therefore I tend to think that the 2nd way you have mentioned is more reasonable.

> 2. For intrinsics that do not specify a specific (N/A) or an agnostic policy, the compiler decides which policy to use. But for policies specified as undisturbed in the intrinsic name, the compiler must be set to be the same.

Preserving the flexibility for compiler optimization when agnostic don't seem like a problem since users expect the elements to be "agnostic".

zakk0610:
> @zakk0610 Another question would like to ask, why put the new intrinsics in the policy_funcs directory instead of merging them together?
In most cases, users would not care about the tail and inactive elements, so I put those policy intrinsic functions into the different folder.

@kito-cheng @zhongjuzhe @nick-knight @eopXD
So all you agree to have `_m_tu`/`_m_ta` for masked reduction and `_mu`/`_ma` for masked compare operation?
any idea?  
I'm asking because I don't like this naming rule so I would like to know your opinion here.
(I still like to use `_tuma`/`_tama` for masked reduction and `_tamu`/`_tama` for masked compare operations. All masked operation always encode the explicitly tail and mask policy regardless operation behavior.  I feel it's clear more because masked always have merge operand.)

zhongjuzhe:
>
I support _m_tu/_m_ta for masked reduction and _mu/_ma for masked compare operation.

kito-cheng:
I like consistent naming rule but I must say `_m_tu`/`_m_ta` is more consistent interface design.

So `_m_tu`/`_m_ta` + 1

nick-knight:
> So all you agree to have `_m_tu`/`_m_ta` for masked reduction and `_mu`/`_ma` for masked compare operation?

Yes, I support both. The naming convention I have in my head is, "If an instruction's behavior is independent of the mask or tail policy setting (`vtype.v{m,t}a`), then the corresponding policy strings (`_{m,t}{u,a}`) are omitted from the corresponding intrinsics names."

I acknowledge that masked reductions are a special case among the (new) "policy" intrinsics since they are the only ones with `_m_t{u,a}` in their names. However, masked reductions are also a special case among RVV instructions since they are the only ones which have masked forms that ignore mask policy. I also acknowledge that the `_m` is reminiscent of the legacy intrinsics. However, we have already agreed on other cases (like `vadc`) where the new API actually overlaps with the legacy API.

nick-knight:
Another option would be to use `vredsum_vs_i8m1_tum` and  `vredsum_vs_i8m1_tam`. This is closer to the nomenclature of other policy intrinsics (`tuma`, etc.).

zakk0610:
Thanks all your opinion! 
I had updated the PR.

lhtin:
> |Masked?  | TU? | MU? | Intrinsic
> |--- |--- |--- |---
> |Yes | No | Yes| `vadd_vv_i8m1_tamu(mask, merge, vs2, vs1, vl)` and `vadd_vv_i8m1_m(mask, merge, vs2, vs1, vl)` if merge is `vundefined()`

@zakk0610 Is it possible to remove vadd_vv_i8m1_m's support for merge as vundefined()? Is it reserved for compatibility reasons?

rvv-intrinsic-rfc.md
```diff
@@ -407,9 +409,9 @@ For reduction intrinsics which may only keep the tail elements.
 |--- |--- |--- |---
 |No  | No | N/A| `vredsum_vs_i8m1_ta(vs1, vs2, vl)`
 |No  | Yes| N/A| `vredsum_vs_i8m1_tu(vd, vs1, vs2, vl)` and `vredsum_vs_i8m1(vd, vs1, vs2, vl)`
-|Yes | No | No | `vredsum_vs_i8m1_tama(mask, vs1, vs2, vl)`
+|Yes | No | No | `vredsum_vs_i8m1_m_ta(mask, vs1, vs2, vl)`
 |Yes | No | Yes| `vredsum_vv_i8m1_m(mask, vd, vs1, vs2, vl)` if vd is `vundefined()`
-|Yes | Yes| No | `vredsum_vs_i8m1_tuma(mask, vd, vs1, vs2, vl)`
+|Yes | Yes| No | `vredsum_vs_i8m1_m_tu(mask, vd, vs1, vs2, vl)`
 |Yes | Yes| Yes| `vredsum_vs_i8m1_m(mask, vd, vs1, vs2, vl)`
```
kito-cheng, line 415: Maybe like this?
```
|Masked?  | TU? | MU? | Intrinsic
|--- |--- |--- |---
|No  | No | N/A | `vredsum_vs_i8m1_ta(vs1, vs2, vl)`
|No  | Yes| N/A | `vredsum_vs_i8m1_tu(vd, vs1, vs2, vl)` and `vredsum_vs_i8m1(vd, vs1, vs2, vl)`
|Yes | No | N/A | `vredsum_vs_i8m1_m_ta(mask, vs1, vs2, vl)`
|Yes | Yes| N/A | `vredsum_vs_i8m1_m_tu(mask, vd, vs1, vs2, vl)` and `vredsum_vs_i8m1_m(mask, vd, vs1, vs2, vl)`
```

kito-cheng: I realized why you write that because we have this in other place
```
|Yes | No | Yes| `vadd_vv_i8m1_tamu(mask, merge, vs2, vs1, vl)` and `vadd_vv_i8m1_m(mask, merge, vs2, vs1, vl)` if merge is `vundefined()`
```

I suggest we could remove this too, and add one note at the end of this section: 
```
NOTE: compiler might change the tail/mask policy to agnostic policy if `merge` or `vd` operand is `vundefined()` value.
```

zakk0610: done. thanks for you suggestion.

eopXD:
Replying to @lhtin 
> Is it possible to remove vadd_vv_i8m1_m's support for merge as vundefined()? Is it reserved for compatibility reasons?

Yes I believe it is reserved for compatibility reasons. When there was no policy function the default behavior, masked function was `tamu` so specifying `vundefined()` was the way to state `ma` behaviors.

lhtin:
> Yes I believe it is reserved for compatibility reasons. When there was no policy function the default behavior, masked function was `tamu` so specifying `vundefined()` was the way to state `ma` behaviors.

Got it, thanks.

zakk0610:
> > I believe the vector register pair passed (by pointer) in `v0` and `v1` serves the role of `undisturbed`.
> 
> @nick-knight Hi, but why does the corresponding mask intrinsics have another dest parameters except for the pointer parameters? For example, `void vlseg2e8_v_i8mf8_m (vint8mf8_t *v0, vint8mf8_t *v1, vbool64_t mask, vint8mf8_t maskedoff0, vint8mf8_t maskedoff1, const int8_t *base, size_t vl);`. I think they should be unified.

Hi @lhtin, are you asking why maksed segment load need to have maskedoff argument? 
IIRC, the original design assumes that tail value is coming from maskedoff argument, not from dest pointer.

rvv-intrinsic-rfc.md
nick-knight: I agree. I will move it.

nick-knight:
In this PR discussion, several concerns have been raised which do not directly relate to the new (mask/tail) policy intrinsics:

- `vmerge` API: https://github.com/riscv-non-isa/rvv-intrinsic-doc/pull/137#discussion_r813524455 . We should probably reopen https://github.com/riscv-non-isa/rvv-intrinsic-doc/issues/126 and have this discussion there.

- masked `vlseg` API: https://github.com/riscv-non-isa/rvv-intrinsic-doc/pull/137#issuecomment-1054226955 . This should be a new issue.

- tail-agnostic `vredsum` API: https://github.com/riscv-non-isa/rvv-intrinsic-doc/pull/137#issuecomment-1049451642 . This should be a new issue. 

In summary, I suggest we do not attempt to address any of these issues in this PR, but rather we define the policy intrinsics preserving the existing APIs in these cases. Then, in separate issues/PRs, we can simultaneously change the old/legacy and the new/policy intrinsics together.

nick-knight:
@zhongjuzhe I believe your comment is relevant to https://github.com/riscv-non-isa/rvv-intrinsic-doc/issues/95

lhtin:
@zakk0610 Can you upload the python script that generates the intrinsics signature together? Because I need it to generate tuple version for segment load and store. Also others can use this to generate intrinsic testcases.

zakk0610:
> Another option would be to use `vredsum_vs_i8m1_tum` and `vredsum_vs_i8m1_tam`. This is closer to the nomenclature of other policy intrinsics (`tuma`, etc.).

I agree to use `_tum` and `_tam` which is closer to the nomenclature of other policy intrinsics.

@lhtin do you mean your gcc would support tuple version segment load/store which difference api with to llvm?

zhongjuzhe:
> 
Yes, our GCC has supported tuple type like vint32m1x2_t. We are actively maintain here: https://github.com/riscv-collab/riscv-gcc/pull/330 as the document evolves. So that would be great if you upload the python script that we can generate tests directly using your script.

zhongjuzhe:
> > Another option would be to use `vredsum_vs_i8m1_tum` and `vredsum_vs_i8m1_tam`. This is closer to the nomenclature of other policy intrinsics (`tuma`, etc.).
> 
> I agree to use `_tum` and `_tam` which is closer to the nomenclature of other policy intrinsics.
> 
> @lhtin do you mean your gcc would support tuple version segment load/store which difference api with to llvm?

Yes, our GCC has supported tuple type like vint32m1x2_t. We are actively maintain here: https://github.com/riscv-collab/riscv-gcc/pull/330 as the document evolves. So that would be great if you upload the python script that we can generate tests directly using your script.

rvv-intrinsic-rfc.md
```diff
@@ -348,70 +348,131 @@ vmerge.vvm vd, vs2, vs1, v0:
...
+|Masked?  | TU? | MU? | Intrinsic
+|--- |--- |--- |---
+|No  | No | N/A| `vadd_vv_<ty>_ta(vs2, vs1, vl)` and `vadd_vv_i8m1(vs2, vs1, vl)`
+|No  | Yes| N/A| `vadd_vv_<ty>_tu(merge,vs2, vs1, vl)`
+|Yes | No | No | `vadd_vv_<ty>_tama(mask, vs2, vs1, vl)`
+|Yes | No | Yes| `vadd_vv_<ty>_tamu(mask, merge, vs2, vs1, vl)`
+|Yes | Yes| No | `vadd_vv_<ty>_tuma(mask, merge, vs2, vs1, vl)`
+|Yes | Yes| Yes| `vadd_vv_<ty>_tumu(mask, merge, vs2, vs1, vl)` and `vadd_vv_i8m1_m(mask, merge, vs2, vs1, vl)`
+
+### Intrinsics with the `dest` argument and doesn't have a masked type
```
eopXD, line 393: Missing a section for "Intrinsics with the `dest` argument and **have** a masked type".

lhtin:
@kito-cheng @zakk0610 @eopXD @nick-knight Is it possible to consider splitting the segment load and store intrinsics into two versions of the API, one using tuple type (e.g. `void vlseg2e8_v_i8mf8_m (vint8mf8_t *v0, vint8mf8_t *v1, vbool64_t mask, vint8mf8_t maskedoff0, vint8mf8_t maskedoff1, const int8_t *base, size_t vl);`, from removed [commit](https://github.com/riscv-non-isa/rvv-intrinsic-doc/blob/19054e8124ac00931eac237c552635989b5d249a/intrinsic_funcs/03_vector_load_store_segment_instructions_zvlsseg.md)) and one existing API(e.g. `void vlseg2e8_v_i8mf8_m (vint8mf8_t *v0, vint8mf8_t *v1, vbool64_t mask, vint8mf8_t maskedoff0, vint8mf8_t maskedoff1, const int8_t *base, size_t vl);`)? The compiler implementer can then choose a version implementation. I known the difficult for LLVM from this [issue 95](https://github.com/riscv-non-isa/rvv-intrinsic-doc/issues/95). However, I hope to provide an option to implement the tuple version to the compiler in the API documentation, if this compiler can implement it.

kito-cheng:
Here is my proposal for the segment load/store API:

Include both API flavor in spec, but mark tuple type style API as optional, and defined some marco to let user check `__riscv_rvv_tuple_type` if compiler has supported.

And non-tuple type style API / new API are mandatory, I believe implement that via tuple type style API like that is feasible, and @rjiejie has implemented this approach in the past https://github.com/riscv-collab/riscv-gcc/pull/307:

```
void vlseg2e8_v_i8mf4 (vint8mf4_t *v0, vint8mf4_t *v1, const int8_t *base, size_t vl)
{
  vint8mf4x2_t vt = vcreate(v0, v1);
  vlseg2e8_v_i8mf4x2 (vt, vase, v0);
}
```

zakk0610:
@kito-cheng @lhtin let's move the tuple type discussion here https://github.com/riscv-non-isa/rvv-intrinsic-doc/issues/139.

zhongjuzhe:
> 

I have implemented the void vlseg2e8_v_i8mf4 (vint8mf4_t *v0, vint8mf4_t *v1, const int8_t *base, size_t vl) using gimple fold inside the GCC in the last commit. We don't need to expand it like this https://github.com/riscv-collab/riscv-gcc/pull/307 in the header. But this api is very ugly I decide to remove. Do you want me to recover that in the this PR https://github.com/riscv-collab/riscv-gcc/pull/334?

zhongjuzhe:
> Here is my proposal for the segment load/store API:
> 
> Include both API flavor in spec, but mark tuple type style API as optional, and defined some marco to let user check `__riscv_rvv_tuple_type` if compiler has supported.
> 
> And non-tuple type style API / new API are mandatory, I believe implement that via tuple type style API like that is feasible, and @rjiejie has implemented this approach in the past [riscv-collab/riscv-gcc#307](https://github.com/riscv-collab/riscv-gcc/pull/307):
> 
> ```
> void vlseg2e8_v_i8mf4 (vint8mf4_t *v0, vint8mf4_t *v1, const int8_t *base, size_t vl)
> {
>   vint8mf4x2_t vt = vcreate_i8mf4x2 (v0, v1);
>   vlseg2e8_v_i8mf4x2 (vt, vase, v0);
> }
> ```

I have implemented the `void vlseg2e8_v_i8mf4 (vint8mf4_t *v0, vint8mf4_t *v1, const int8_t *base, size_t vl)` using gimple fold inside the GCC in the last commit. We don't need to expand it like this https://github.com/riscv-collab/riscv-gcc/pull/307 in the header. But this api is very ugly I decided to remove. Do you want me to recover that in the this PR https://github.com/riscv-collab/riscv-gcc/pull/334? I am OK with that.

kito-cheng:
> I have implemented the void vlseg2e8_v_i8mf4 (vint8mf4_t *v0, vint8mf4_t *v1, const int8_t *base, size_t vl) using gimple fold inside the GCC in the last commit. We don't need to expand it like this https://github.com/riscv-collab/riscv-gcc/pull/307 in the header. But this api is very ugly I decided to remove. Do you want me to recover that in the this PR https://github.com/riscv-collab/riscv-gcc/pull/334? I am OK with that.

Personally I don't like that API too, but it's would be great to having consistent API between LLVM and GCC, it's compromise on the API design part, so yes, plz add back into GCC implementation.

rvv-intrinsic-rfc.md
zakk0610: In fact, Intrinsics with the `vd` argument is the same as "Intrinsics with the dest argument and have a masked type" because we call it `vd` rather than `dest`. https://github.com/riscv-non-isa/rvv-intrinsic-doc/pull/104

zhongjuzhe:
> > I have implemented the void vlseg2e8_v_i8mf4 (vint8mf4_t *v0, vint8mf4_t *v1, const int8_t *base, size_t vl) using gimple fold inside the GCC in the last commit. We don't need to expand it like this [riscv-collab/riscv-gcc#307](https://github.com/riscv-collab/riscv-gcc/pull/307) in the header. But this api is very ugly I decided to remove. Do you want me to recover that in the this PR [riscv-collab/riscv-gcc#334](https://github.com/riscv-collab/riscv-gcc/pull/334)? I am OK with that.
> 
> Personally I don't like that API too, but it's would be great to having consistent API between LLVM and GCC, it's compromise on the API design part, so yes, plz add back into GCC implementation.

I have implemented the nontuple segment load store and add fully tested testcases now. Can you review that for me? The most important implementation are in gcc/config/riscv/riscv-vector-builtins-functions.cc.  I used gimple fold the intrinsic api.

zhongjuzhe:
> Support the policy intrinsic functions to control tail and mask policy.
> 
> Some operations would not take tail and mask value, so they would not have policy intrinsic functions. So in this proposal, I would like to keep the current policy rule for all non-policy intrinsic functions (ex. vadd_vv_i8m1 and vadd_vv_i8m1_m). It is the only one different thing comparing with proposal in #120
> 
> Please see [rvv-intrinsic-rfc.md ](https://github.com/riscv-non-isa/rvv-intrinsic-doc/blob/new-policy-proposal/rvv-intrinsic-rfc.md#policy-intrinsic-functions) for more detail.
> 
> There are some special cases list as below:
> 
> 1. vadc/vsbc support `_ta` and _`tu`.
> 2. vmadc/vmsbc support `_ta`.
> 3. Vector Integer and Floating-Point Comparison Instructions support `_ta`, `_tamu` and `_tama`.
> 4. Vector Integer and Floating-Point Move support `_ta` and `_tu`.
> 5. vlm.v support `_ta`.
> 6. vmsbf.m/vmsif.m/vmsof.m support `_ta`, `_tama` and `_tamu`.
> 7. vmv.s.x/vfmv.s.f support `_ta` and `_tu`.
> 8. Vector Integer and Floating-Point Merge Instructions support `_ta` and `_tu`.
> 9. Vector Compress Instruction support `_ta` and `_tu`.
> 10. Vector Store, vfirst.m, vmv.x.s, vfmv.f.s, vcpop.m and Miscellaneous Functions don't have don't have policy intrinsic functions.
> 11. vslideup and vslidedown don't have dest operand in `_ta` and `_tama`.
> 
> For overloading functions, encode the policy in the end of function name. For example, vadd would have `vadd_ta`, `vadd_tu`, `vadd_tama`, `vadd_tamu`, `vadd_tuma` and `vadd_tumu`.

Are you planning to merge this proposal into master branch? I have implemented all intrinsics in your proposal.

kito-cheng:
@lhtin 
> @zakk0610 Can you upload the python script that generates the intrinsics signature together? Because I need it to generate tuple version for segment load and store. Also others can use this to generate intrinsic testcases.

We've plan on release that, and consulting with RVI folks about the licence and copyright issue now.

lhtin:
> We've plan on release that, and consulting with RVI folks about the licence and copyright issue now.

Thank you so much @kito-cheng.

zakk0610:
>> Another option would be to use vredsum_vs_i8m1_tum and vredsum_vs_i8m1_tam. This is closer to the nomenclature of other policy intrinsics (tuma, etc.).

> I agree to use _tum and _tam which is closer to the nomenclature of other policy intrinsics.

> Are you planning to merge this proposal into master branch? I have implemented all intrinsics in your proposal.
@zhongjuzhe awesome!  
Sorry, It seems to using  `_tum` and `_tam` is closer to the nomenclature of other policy intrinsics. I would make the latest changed for this today and wait a while to see any other suggestions.

zhongjuzhe:
> 

OK. When will you merge this propsal into master branch? And any change? I would like to collect all changes and change implementation in GCC together?

zhongjuzhe:
> > > Another option would be to use vredsum_vs_i8m1_tum and vredsum_vs_i8m1_tam. This is closer to the nomenclature of other policy intrinsics (tuma, etc.).
> 
> > I agree to use _tum and _tam which is closer to the nomenclature of other policy intrinsics.
> 
> > Are you planning to merge this proposal into master branch? I have implemented all intrinsics in your proposal.
> > @zhongjuzhe awesome!
> > Sorry, It seems to using  `_tum` and `_tam` is closer to the nomenclature of other policy intrinsics. I would make the latest changed for this today and wait a while to see any other suggestions.

OK. When will you merge this propsal into master branch? And any other changes? I would like to collect all changes and change implementation in GCC together?

zakk0610:
Currently that's all we have.
It's big set of intrinsics we need to support, I think maybe let we wait a more week to see any more opinions?

@kuanlinchentw @rofirrim any concern?

@kito-cheng do you know who would intereset in this?

zhongjuzhe:
> 

OK, I am gonna wait until the this proposal is stable. Then I change the GCC implementation. By the way, will you announce the intrinsic api document is frozen in the future? Just like the spec of RVV.

zhongjuzhe:
> Currently that's all we have. It's big set of intrinsics we need to support, I think maybe let we wait a more week to see any more opinions?
> 
> @kuanlinchentw @rofirrim any concern?
> 
> @kito-cheng do you know who would intereset in this?

OK, I am gonna wait until the this proposal is stable. Then I change the GCC implementation. By the way, will you announce the intrinsic api document is frozen in the future? Just like the spec of RVV.

lhtin:
> Sorry, It seems to using _tum and _tam is closer to the nomenclature of other policy intrinsics. I would make the latest changed for this today and wait a while to see any other suggestions.

If so, should `_ma` be changed to `_tma`?

zakk0610:
> By the way, will you announce the intrinsic api document is frozen in the future? Just like the spec of RVV.

I think we need to discuss this in another issue thread. 

> If so, should _ma be changed to _tma?
Thanks, It makes sense to me, I will update document later if no other say no.

zhongjuzhe:
> > By the way, will you announce the intrinsic api document is frozen in the future? Just like the spec of RVV.
> 
> I think we need to discuss this in another issue thread.
> 
> > If so, should _ma be changed to _tma?
> > Thanks, It makes sense to me, I will update document later if no other say no.

Are you agree to change `_ma` to `tma`? just like `tu` to `tum`? I think for the consistency, it should be.

issuehsu:
I agree with Nick's viewpoint.

- I think we are on the same page for the masked reduction operation
    - _m_tu/_m_ta ==> _tum/_tam
    - the tail m not only means we don't care mu/ma, but it also implies it is a mask operation. So we need to keep tail m.
- For masked compare operation
    - _mu/_ma ==> _tma _tmu
    - the _t means we don't care tu/ta, and has no further implication. It is irrelevant and could be skipped.

lhtin:
I just think `_tma` and `_tmu` are more unified with `_txmx`(x can be `a`, `u`, `ε`(mean empty)) nomenclature than `_ma` and `_mu`.

zhongjuzhe:
I am agree with the Nick's viewpoint now. We can keep _mu/_ma instead of _tma/_tmu. What I care is that when will you merge these new API intrinsics and freeze the RVV C api intrinsics? Because I am planning to push GCC implementation uptream recently and I think I should wait until these new API C intrinsics be stable. Thanks~

kito-cheng:
@zhongjuzhe Apology for the unclear schedule for the API freeze, we'll have a schedule for the API development and publish that soon.

zhongjuzhe:
> @zhongjuzhe Apology for the unclear schedule for the API freeze, we'll have a schedule for the API development and publish that soon.

It's OK. Our company's FSF just has been approved yesteday. I start to think about how to push codes upstream and trying to configure out the flow of contributing codes using git email~.

zhongjuzhe:
I have a question related to overloaded functions. I see `vint8mf8_t vadd_tu (vint8mf8_t merge, vint8mf8_t op1, vint8mf8_t op2, size_t vl);` in the new api document. However, I thinks the suffix `_tu` is not necessary. Because the argument `merge` can be used to classify the `ta` or `tu` as follows:
1.  vint8mf8_t vadd (vint8mf8_t op1, vint8mf8_t op2, size_t vl); generate `ta` policy.
2.  vint8mf8_t vadd (**vint8mf8_t **merge****,  vint8mf8_t op1, vint8mf8_t op2, size_t vl); generate `tu` policy.

zakk0610:
In some case like vmacc still have the same interface, IMO, I prefer to keep `_ta` and `_tu` to make the consistent for all policy functions. 

```
vint8mf8_t vmacc_vv_i8mf8_tu (vint8mf8_t vd, vint8mf8_t vs1, vint8mf8_t vs2, size_t vl);
vint8mf8_t vmacc_vv_i8mf8_ta (vint8mf8_t vd, vint8mf8_t vs1, vint8mf8_t vs2, size_t vl);
```

zhongjuzhe:
OK, for api consistency. Let's keep it. By the way, will these api be merged into `master` branch?

kito-cheng (approved):
LGTM, we've PoC on LLVM by @zakk0610 and GCC implementation by @zhongjuzhe, so I believe it's ready to merge, thanks everyone!

PR #138, opened 2022-03-04
Adapt adoc style and revamp the API reference manual

eopXD:
As the v-spec is now ratified and the existing RVV intrinsics are less likely to change, I think it is a good time to start and revamp the API reference manual. Adapting it to `.adoc` just like v-spec allows it to be compiled into pdf so the RVV intrinsic functions can be formally released in the future.

When adapting to `adoc` style, I also tried to do the following:

- Append clarifications and details under `rvv-intrinsic-rfc.md` that documented design decisions of the intrinsics them into the API reference manual
- Add some descriptions when listing out the functions rather than just enumerating the possible prototypes

Please take a look into [rvv-intrinsic-api.adoc](https://github.com/eopXD/rvv-intrinsic-doc/blob/improve-api-doc/rvv-intrinsic-api.adoc) and compare them to the old ones [rvv-intrinsic-api.md](https://github.com/riscv-non-isa/rvv-intrinsic-doc/blob/master/rvv-intrinsic-api.md)


eopXD:
Dropping this and opening another one.

PR #143, opened 2022-03-17
Modify return value in example for coherence

8U61ife:


examples/rvv_memcpy.c
```diff
@@ -27,4 +29,4 @@ int main() {
 
   // compare
   puts(compare_1d(golden, actual, N) ? "pass" : "fail");
-}
```
eopXD, line 30: May I ask why is this line updated?

eopXD (approved):
LGTM except the minor comment.

examples/rvv_memcpy.c
8U61ife: In fact, I don't know.I just use vim to edit.

8U61ife: It seems that there is a newline character which auto apeend in my file. Although I try to press `delete` at the end.

eopXD: Please remove it as this is not related to the topic of this PR. Coding style issues shall be dealt in a separately.

8U61ife: Done.

eopXD:
Thanks for the PR.

Issue #48, opened 2020-11-17
Remove splats?


nick-knight:
The `v{f}splat` routines appear to just be simple aliases for the `v{,f}mv_v_{x,f}` routines. There's obviously no harm in including them; however, in such an enormous API I think we should take measures to reduce complexity whenever possible. Thus I propose removing the splats.

Less drastic measures are (1) to deprecate them now and revisit removing them sometime in the future, or (2) to make them an optional extension.

knightsifive:
Upon further study, it appears that all of the "semantic functions"
https://github.com/riscv/rvv-intrinsic-doc/blob/master/rvv_intrinsic_funcs/13_semantic_vector_functions.md
are redundant. Hmm. Obviously this was intentional, meaning that someone out there clearly wanted these aliases.

I guess these could all be implemented as macro wrappers around existing routines... Another idea is that we just formally define these routines as such, and provide a separate header file that implements these translations.

HanKuanChen:
I think the reason we have semantic functions is that the intrinsic document does not use [v-spec](https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc) naming in the [beginning](https://github.com/riscv/rvv-intrinsic-doc/blob/2be63fc207f239f1d0c6691848f4d78523e6b985/rvv_intrinsic_funcs.md). Provide semantic functions help users programming easily.
However, after upgrading to [new interface](https://github.com/riscv/rvv-intrinsic-doc/blob/bb210940fa9b7c9c4e204c6cade7e901860349a0/rvv_intrinsic_funcs.md), the semantic functions somehow remain.

knightsifive:
Since they're vestigial and now redundant, I'm casting my vote in favor of excising them.

rednoah91:
From my using experience, I also vote for removing splats.

zakk0610:
It seems all we agree remove splat.
Another question is do we really need the "semantic functions"?
Personally I prefer to remove all "semantic functions" in intrinsic rfc.

zakk0610:
currently I remove splat function in https://github.com/riscv/rvv-intrinsic-doc/commit/17d4cb460cab39d1bb567af085ee493b74bd26b8.
Is ok to remove all "semantic functions" ?

HanKuanChen:
I think our opinion is remove whole semantic functions.

eopXD:
We don't have any "semantic functions" now. Closing this issue.

PR #146, opened 2022-03-23
Clean-up intrinsics

eopXD:
This PR does the following:

- Remove Zvamo: Zvamo is removed in v-spec v1.0 and the RVV C intrinsic needs to follow.
- Remove current policy intrinsic: The policy intrinsics were introduced in #116. The policy intrinsics' interface is still under discussion and neither of the upstream compilers (LLVM or GCC) support them. Therefore we remove them and wait until the on-going discussions (#137) converge to a new one.
- Remove "semantic functions": The discussion in #48 came to a conclusion that they are all redundant and needs to be removed. This PR addresses it.


zakk0610 (approved):
LGTM.

kito-cheng (approved):

PR #155, opened 2022-07-29
Fix broken links

topperc:
The file names were renumbered at some point

Signed-off-by: Craig Topper <craig.topper@sifive.com>


eopXD (approved):
LGTM, thank you!

PR #161, opened 2022-09-07
Attempt to repair links from rvv-intrinsics-api.md

topperc:
Signed-off-by: Craig Topper <craig.topper@sifive.com>


eopXD (approved):
Looks good. Hopefully this kind of maintenance can go away when we switch to ASCII Doc.

PR #172, opened 2022-10-08
Replace vpopc with vcpop in exmaple/rvv_reduce.c

msyksphinz-self:
RVV1.0, `vpopc.m` is replaced into `vcpop.m`. riscv-gnu-toolchain latest assembler generated error.

This PR replace example code `vpopc.m` into `vcpop.m` for RVV1.0


eopXD (commented):
Thank you for the PR. The v-spec 1v.0 now has `vcpop.m rd, vs2, vm` to be aligned with its scalar version but also keeps the previous naming `vpopc` as it's alias.

On the other hand, your PR guided me to discover that the current `intrinsic.md` only documents `vcpop` and didn't include the alias naming `vpopc`. I'll look into it and also possible cases, thanks.

If possible, please add some descriptions in the commit message and a line of`Signed-off by: Author Name <author@mail.com>`.

eopXD (approved):

PR #173, opened 2022-10-13
Add macro to imply compiler availability on intrinsic versions

eopXD:
The approach to API change will involve multiple releases, so we need a macro to imply the version supported in the compiler.

The value calculation of version number is the same as [riscv-non-isa/riscv-c-api-doc](https://github.com/riscv-non-isa/riscv-c-api-doc).

Adding the document as ASCIIDoc format since we will soon re-format other markdowns to ASCIIDoc as well. The document will be restructured and not be a single document like this PR is proposing right now.


kito-cheng (approved):
LGTM

rofirrim (approved):
LGTM. Thanks!

Issue #180, opened 2022-12-27
Exposure of vlenb missing for vread_csr


eopXD:
A frequently asked question about the API is why aren't vlenb available. It should also be something exposed by the intrinsic, this issue is a placeholder of this and we should resolve it.

tingsi5:
Is it possible to take `vl` and `vtype` into consdieration as well?

nick-knight:
> Is it possible to take vl and vtype into consdieration as well?

Can you describe the use-case you envision?

My opinion is we should maintain the current API where the `vl` and `vtype` CSRs are abstracted away from the C programmer. In the current API, the user specifies the desired semantics with the `size_t vl` argument and suffixes like `i64m8_tumu` in the intrinsics names, leaving the compiler to generate the necessary CSR accesses. Since the compiler has the freedom to reschedule code (as long as it preserves semantics), the C programmer will generally not know where, in the C program order, these accesses will be generated, so it's hard for me to envision how a programmer would use this mechanism usefully.

Also, adding more visibility into `vl` and `vtype` would take us down a similar path to C `fenv`, which is something we've been trying to avoid from the beginning of this project. I'm happy to go into more details.

`vlenb` is a different scenario because it is read-only (moreover, a "design-time constant"). So the underlying CSR read can be safely generated independently of surrounding (vector) code-gen

tingsi5:
Now I understand. Thank you for explaining me the reason why `vl` and `vtype` are not included from the beginning. I asked because as an user, it's unclear for me why read-only CSRs are excluded from the read API and thought if we're going to add `vlenb`, then probably we can take `vl` and `vtype` as well (I expected those CSRs are exposed intentionlly for read and should be included in the API).

If you have time I'd like to know the story of `fenv`. I googled it but got only the documentation.

PR #182, opened 2023-01-05
Expose vlenb to user

eopXD:
This resolves riscv-non-isa/rvv-intrinsic-doc#180.

Corresponding LLVM change is [D141032 [Clang][RISCV] Expose vlenb to user](https://reviews.llvm.org/D141032).

### DIFF
--- a/rvv-intrinsic-api.md
+++ b/rvv-intrinsic-api.md
@@ -70,6 +70,8 @@ enum RVV_CSR {
 
 unsigned long vread_csr(enum RVV_CSR csr);
 void vwrite_csr(enum RVV_CSR csr, unsigned long value);
+
+unsigned long vlenb();
 ```
 
 ## 7. Vector Loads and Stores


eopXD:
Change: Open up a separate function for access to vlenb.

Upon a second thought, with future extensions of the intrinsics, we will add rounding mode and exception intrinsics. The exposure of `vxsat`, `vxrm`, and `vcsr` in `vread_csr` and `vwrite_csr` will confuse user and we should remove them. Furthermore, `vstart` is pretty much always kept to zero, and intrinsic users probably should not have access to them.

With the considerations above, I think `vread_csr` and `vwrite_csr` can be removed in the future when the rounding mode and exception intrinsics are added. This is why this PR is opening up a separate function to provide access to `vlenb`.

eopXD:
With concern for naming collision for such short name, I believe it will be resolved when we add a common suffix to all intrinsics in the future.

[0] On-going work for a common prefix: https://github.com/riscv-non-isa/riscv-c-api-doc/pull/31

nick-knight (approved):
LGTM. Agree with your comment that we should remove `v{read,write}_csr` and that this can be done in a follow-up PR.

rofirrim (approved):
LGTM. Thanks @eopXD !

PR #186, opened 2023-01-11
Simplify the intrinsics by blending non-policy / policy intrinsics into a single set

eopXD:
## Intro

This works implements the simplification proposal proposed by Nick Knight @nick-knight [0]. The simplification proposal targets to blend the two existing sets of intrinsics, non-policy / policy intrinsics, into a single set, with default assumption of policy behavior as TAMA. The simplification reduces 25% of the overall out-going intrinsics.

This pull request can be split into 3 parts.
- Change default TUMU intrinsics to TUMA
- Change all intrinsics from TAMU to TAMA
- Simplify intrinsics with policy suffix

The recommended way of reviewing this patch is to checkout the changes under each instruction mnemonics. Take `vadd` for example, you can checkout `auto-generated/api-testing/vadd.c`, `auto-generated/overloaded-api-testing/vadd.c`, `auto-generated/policy_funcs/llvm-api-tests/vadd.c`, `auto-generated/policy_funcs/llvm-overloaded-tests/vadd.c` to see the prototype changes. 

[0] https://gist.github.com/nick-knight/6cb0b74b351a25323dfb1821d3a269b9

### For intrinsics originally with default as TAMU

Before:

```cpp
vint32m1_t vadd_vv_i32m1 (vint32m1_t op1, vint32m1_t op2, size_t vl);
vint32m1_t vadd_vv_i32m1_m (vbool32_t mask, vint32m1_t maskedoff, vint32m1_t op1, vint32m1_t op2, size_t vl);
vint32m1_t vadd_vv_i32m1_tu (vint32m1_t merge, vint32m1_t op1, vint32m1_t op2, size_t vl);
vint32m1_t vadd_vv_i32m1_ta (vint32m1_t op1, vint32m1_t op2, size_t vl);
vint32m1_t vadd_vv_i32m1_tuma (vbool32_t mask, vint32m1_t merge, vint32m1_t op1, vint32m1_t op2, size_t vl);
vint32m1_t vadd_vv_i32m1_tumu (vbool32_t mask, vint32m1_t merge, vint32m1_t op1, vint32m1_t op2, size_t vl);
vint32m1_t vadd_vv_i32m1_tama (vbool32_t mask, vint32m1_t op1, vint32m1_t op2, size_t vl);
vint32m1_t vadd_vv_i32m1_tamu (vbool32_t mask, vint32m1_t merge, vint32m1_t op1, vint32m1_t op2, size_t vl);
```

After:

```cpp
[...]

PR #188, opened 2023-01-13
Change default policy of intrinsics without a policy suffix from TAMU to TAMA

eopXD:
This PR is currently dependent to #186.

---

This works is one of the steps to implement the simplification proposal
proposed by Nick Knight [0]. After setting the default policy to TAMA,
the next-step will be to remove the `_ta` and `_tama` variant and 
further simply other policy suffixes.

[0] https://gist.github.com/nick-knight/6cb0b74b351a25323dfb1821d3a269b9

The `maskedoff` operand is removed for masked version of the non-policy
intrinsics in this PR.


eopXD:
Closing the PR, after a second thought, I think having changes in a single PR would be more sensible.

PR #192, opened 2023-02-08
rvv examples improvement

XiaoWang1772:
- examples/rvv_strcpy: increase copy length
    According to sections 15.4 and 15.5 of the riscv-v-spec-1.0, vmsbf and
    vmsif will set 1 to all the active mask elements whose indexes are smaller
    (or <= for vmsif) than the first active source element that is a 1, a
    special case of these instructions is when the source operand is all-0s,
    then the destination register would be set to all-1s.

    This special case is important for long string copy where the terminator 0
    won't be detected in the first iteration, so this patch increases the
    string length from 100B to 2000B to cover the case. 2000B should be big
    enough to cover the mainstream VLEN implementations with LMUL set to 8.

- examples/rvv_reduce: remove unused variable


eopXD (approved):
LGTM, thank you for the contribution!

PR #195, opened 2023-02-13
Add --skip-default-inst option to skip rendering of the RVV instructions

eopXD:
This is a part of the feature to allow other vendors to hook their own instructions (or generator) to this framework. A full example of the usage will be added in the future.


kito-cheng (approved):
LGTM

PR #196, opened 2023-02-13
Fix filename for API testing

eopXD:
The filenames was originally parsed with code like the following, probably a less maintainable way we want.

```
test_file_name = f"{overloaded_func_name.split('_')[3]}.c"
```

This PR fixes this by passing the actual mnemonic as a parameter and assigning it as the filename. Files are deleted / added due to this change, but the actual contents of the unit-test don't change.


eopXD:
Landing this implies to adjust the test cases under llvm-project too to keep them in-sync.

kito-cheng (commented):
Maybe encoding that info to inst_info?

eopXD:
> Maybe encoding that info to inst_info?

Good point, changed 1st commit to reuse mnemonic already recorded under `inst_info`.

kito-cheng (approved):
LGTM

PR #201, opened 2023-02-20
Clarify semantics for scalar operands

### LINKED ISSUE #199: Question about "If XLEN < SEW, the value from the scalar register is sign-extended to SEW bits."
hi，
"If XLEN < SEW, the value from the scalar register is sign-extended to SEW bits" is described in the [documen](https://github.com/riscv-non-isa/rvv-intrinsic-doc/blob/5818330c298306714fefb705a524ddf3b9b50125/rvv-intrinsic-rfc.md#scalar-in-vector-operations)t, and I have some questions about it:

The function ```int64_t vmv_x_s_i64m1_i64 (vint64m1_t a);``` on RV32, that is, XLEN = 32,SEW = 64, has two situations:

1. According to the description "If XLEN < SEW, the value from the scalar register is sign-extended to SEW bits", only the lower 32 bits of ```v1``` register should be intercepted into a 32-bit scalar register ```a0```, and then ```a0``` should be signed and extended to 64 bits (```a1:a0```).

2. According to my understanding, the lower 32 bits of ```v1``` register should be copied to a 32-bit scalar register ```a0```, then the v1 register should be shifted to the right by 32 bits to generate vector register ```v2```, then the lower 32 bits of v2 should be copied to a 32-bit scalar register ```a1```, and finally 64-bit data (```a1:a0```)

I think 2 is more appropriate, because 1 will discard the high 32-bit data of ```v1```, but that "If XLEN < SEW, the value from the scalar register is sign-extended to SEW bits" makes me confused. Can someone answer for me? 
Thank you very much!

nick-knight:
The existing text was phrased in a way that may cause confusion in the case of instructions with scalar outputs, like `vmv.x.s`. 

There was also some redundancy, a typo in one of the assembly mnemonics, and failure to mention that the scalar operands are sometimes zero- (vs. sign-) extended.

I've rewritten the section to address these things, and close #199.

### DIFF
--- a/rvv-intrinsic-rfc.md
+++ b/rvv-intrinsic-rfc.md
@@ -256,7 +256,7 @@ vuint64m1_t vadd_vx_u64m1(vuint64m1_t op1, uint64_t op2, size_t vl);
 
 To support the implied semantics, the compiler may generate multiple instructions for the intrinsics.
 (More generally, this API does not constrain the compiler in which instructions it actually generates.)
-For example, to support `vmv_x_s_i64m1_i64` when XLEN = 32, the compiler may generate `vmv.x.s`, `vslide1down.vi`, and another `vmv.x.s`, with all SEW = 32, to extract the full 64 bits into two X-registers. (A future extension of the intrinsics could include an "`vmv_x_s_i64m1_i32`", perhaps only when XLEN = 32, whose semantics match `vmv.x.s` for XLEN = 32 and SEW = 64.)
+For example, to support `vmv_x_s_i64m1_i64` when XLEN = 32, the compiler may generate `vmv.x.s`, `vslide1down.vi`, and another `vmv.x.s`, all with SEW = 32, to extract the full 64 bits into two X-registers. (A future API extension could include an "`vmv_x_s_i64m1_i32`", perhaps only when XLEN = 32, whose semantics match `vmv.x.s` for XLEN = 32 and SEW = 64.)
 
 ## Mask in Intrinsics<a name="mask-in-intrinsics"></a>
 
--- a/rvv-intrinsic-rfc.md
+++ b/rvv-intrinsic-rfc.md
@@ -257,7 +257,6 @@ vuint64m1_t vadd_vx_u64m1(vuint64m1_t op1, uint64_t op2, size_t vl);
 To support the implied semantics, the compiler may generate multiple instructions for the intrinsics.
 (More generally, this API does not constrain the compiler in which instructions it actually generates.)
 For example, to support `vmv_x_s_i64m1_i64` when XLEN = 32, the compiler may generate `vmv.x.s`, `vslide1down.vx`, and another `vmv.x.s`, all with SEW = 32, to extract the full 64 bits into two X-registers.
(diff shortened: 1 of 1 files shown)


nick-knight:
Originally I removed references to floating-point from this section, because I thought it was irrelevant. Indeed, the API follows the underlying ISA extensions by precluding SEW > FLEN. On the other hand, SEW < FLEN is possible and handled similarly to the integer case. I've just pushed a minor tweak to cover this.

rvv-intrinsic-rfc.md
```diff
@@ -254,18 +254,9 @@ vuint8m1_t vadd_vx_u8m1(vuint8m1_t op1, uint8_t op2, size_t vl);
...
-
-```
-Example:
-
-// Use uint8_t for op2.
-vuint8m1_t vslide1up_vx_u8m1(vuint8m1_t dest, vuint8m1_t op1, uint8_1 op2);
-```
+To support the implied semantics, the compiler may generate multiple instructions for the intrinsics.
+(More generally, this API does not constrain the compiler in which instructions it actually generates.)
+For example, to support `vmv_x_s_i64m1_i64` when XLEN = 32, the compiler may generate `vmv.x.s`, `vslide1down.vi`, and another `vmv.x.s`, all with SEW = 32, to extract the full 64 bits into two X-registers. (A future API extension could include an "`vmv_x_s_i64m1_i32`", perhaps only when XLEN = 32, whose semantics match `vmv.x.s` for XLEN = 32 and SEW = 64.)
```
topperc, line 259: Probably a question for @eopXD, but should the intrinsics in this file use `__riscv_` prefixed names?

topperc: FWIW, clang will drop the extra instructions if the result is stored to a 32-bit type.

nick-knight: Let's rename things in a separate PR. 

Thanks for pointing out LLVM's optimization. I didn't know that and it makes me feel better about this "omission". However, I'm not sure it needs to be mentioned in the API doc.

rvv-intrinsic-rfc.md
```diff
@@ -254,18 +254,10 @@ vuint8m1_t vadd_vx_u8m1(vuint8m1_t op1, uint8_t op2, size_t vl);
...
-```
-Example:
-
-// Use uint8_t for op2.
-vuint8m1_t vslide1up_vx_u8m1(vuint8m1_t dest, vuint8m1_t op1, uint8_1 op2);
-```
+To support the implied semantics, the compiler may generate multiple instructions for the intrinsics.
+(More generally, this API does not constrain the compiler in which instructions it actually generates.)
+For example, to support `vmv_x_s_i64m1_i64` when XLEN = 32, the compiler may generate `vmv.x.s`, `vslide1down.vx`, and another `vmv.x.s`, all with SEW = 32, to extract the full 64 bits into two X-registers.
+(A future API extension could include an "`vmv_x_s_i64m1_i32`", perhaps only when XLEN = 32, whose semantics match `vmv.x.s` for XLEN = 32 and SEW = 64.)
```
eopXD, line 260: I think it will be better if we track this under `Issue`, or else this idea may get lost in this rfc file.

eopXD (approved):
Looks good to me except the single comment. Thank you for clarifying this Nick :)

PR #202, opened 2023-02-21
Fix details of the compatible header

eopXD:
This PR adds missing details of the compatible header.

- Add brackets to the macro to guard against unintended behavior.
- Add include guard for the compatible header when this header is shipped into the toolchain next to the existing `riscv_vector.h`
- Fix policy mapping for the legacy non-overloaded intrinsics (according to spec. it was supposed to be `tamu`, but the compiler actually implemented a more strict circumstance - `tumu`)
- Amend edges cases of the suffix (where instructions don't have `mu` variant, or is always `ta`).


eopXD:
Gentle ping, thank you.

nick-knight (approved):
LGTM.

eopXD:
The PR description is updated. The header is verified with `make run-non-policy-compatible-api-testing`.

PR #221, opened 2023-04-24
Define reinterpret intrinsics between vector boolean types and LMUL=1 (m1) vector integer types

### LINKED ISSUE #156: [RFC] Use case for type casting of boolean vector values
Issue in #69 points out that current vector mask registers all follow MLEN = 1, which bit values are compact inside a single vector register ([4.5. Mask Register Layout](https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc#sec-mask-register-layout)). Current bool types exist because of type safe reasons. 

However the issue implies that there may be some use case on type casting these boolean vector registers. We want to ask for more input on this before we decide to add more intrinsics API-s.

- What are the use cases where we want type conversions between different `vbool` values?
- What are the use cases of casting between boolean vector registers to LMUL<=1 values integer values?

eopXD:
Resolves #156.

The LLVM implementation is
D149642 [[RISCV] Support vreinterpret intrinsics between vector boolean type and m1 vector integer type](https://reviews.llvm.org/D149642)


rvv-intrinsic-generator/rvv_intrinsic_gen/templates/reint_op_template.py
```diff
@@ -112,4 +113,43 @@ def render(G, op_list, type_list, sew_list, lmul_list, decorator_list):
...
+    G.write("// Reinterpret between vector boolean types and LMUL=1 (m1)")
+    G.write(" vector integer types\n")
+    # [integer type, integer short type]
+    convert_set = [["int", "i"], ["uint", "u"]]
+    for args in prod(
+        OP=op_list, SEW=sew_list, TYPES=convert_set, LMUL=lmul_list):
+
+      args["OP"] = "v" + args["OP"]
+
+      type_helper = TypeHelper(**args)
```
rofirrim, line 130: This is just a comment, no action needed.

I was initially confused by `LMUL` not apparently being used in this loop but `TypeHelper` does use it to compute the right `MLEN` type.

rofirrim (approved):
LGTM. Thanks a lot @eopXD!

rvv-intrinsic-rfc.md
```diff
@@ -566,6 +567,19 @@ Example:
 vint64m1_t vreinterpret_v_i32m1_i64m1(vint32m1_t src)
 ```
 
+### Reinterpret between vector boolean types and LMUL=1 (m1) vector integer types<a name="reinterpret-vbool-o-vintm1"></a>
+
+These utilities help users to convert vector boolean types to `LMUL=1` vector integer types, e.g. converting `vbool64_t` to `vint64m1_t`, `vbool16_t` to `vint8m1_t`. The reinterpret intrinsics only change the types of the underlying contents. It is a nop operation.
```
topperc, line 572: What are the semantics of the extra bits that don't exist in the bool type when converting to the int type?

eopXD: The reinterpret function is a pseudo function that lets user goes over the type safe scheme boundary.

I think I was wrong in this paragraph.  Currently the v-spec specifies `v0` as the single supported mask register. The semantic of the reinterpret cast shall be a `vmv.v.v` between the mask register (`v0`) and the `m1` integer register (which can be anything in `v8` ~ v32`). Revised the paragraph.

rvv-intrinsic-generator/rvv_intrinsic_gen/templates/reint_op_template.py
eopXD: Yes, you are correct.

rvv-intrinsic-rfc.md
topperc: v0 is the only _input_ mask register. vmseq and similar can write any register with mask data.

The vbool64_t type only contains vlen/64 bits. Casting it to m1 leaves other (63/64)*vlen bits undefined. As far as the compiler is concerned we'd be allowed to have them be any value. It would be difficult for the user to control. I supposed a lmul_ext has the same issue so maybe this is expected?

eopXD: I was confused by myself in the previous comment, please disregard them.

I think the user should be aware of the mask register layout and know that the remaining bits are undefined.

eopXD: Added another note in [5bcf9a4](https://github.com/riscv-non-isa/rvv-intrinsic-doc/pull/221/commits/5bcf9a49f912c48ace1aea15cd10e2dec10af8a6) to aware users of the mask register layout.

jan-wassenberg: hm, is there some other way to go from vbool64_t to vbool1_t? Otherwise casting still seems important, even if some of the bits are undefined.

kito-cheng: That still possible by doing something like this: `vbool64_t` -> `m1` -> `vbool1_t`.

jan-wassenberg: Ah, perhaps I misunderstood, it sounded like there was a proposal not to support vbool64_t -> m1. I think that would be too limited.

eopXD:
Merging this upon approval and given that LLVM has a patch ready (mentioned in the PR's description).

Incarnation-p-lee:
Thanks @eopXD . Just FYI, the GCC upstream has implemented this VREINTERPRET intrinsic already.

PR #222, opened 2023-04-25
Model rounding mode control for fixed-point intrinsics

eopXD:
This PR is based on previous [RFC post](https://gist.github.com/eopXD/f1001bf4745a44815536fd2ac59decc7), this replaces the existing fixed-point intrinsics and adds new fixed-point intrinsics with prefix implying the control to the fixed-point rounding mode register `vxrm`.

Edit: Through discussion below, the current PR is not going to static encode the rounding mode into the function name and is using an parameter that is restricted for constant inputs only.

```cpp
- vint32m1_t __riscv_vsadd_vv_i32m1 (vint32m1_t op1, vint32m1_t op2, size_t vl);
+ vint32m1_t __riscv_vsadd_vv_i32m1 (vint32m1_t op1, vint32m1_t op2, unsigned int vxrm, size_t vl);
```

The LLVM implementation are the following:

- D151395 [[1/3][RISCV] Define machine instruction to write an immediate into vxrm](https://reviews.llvm.org/D151395)
- D151396 [[2/3][RISCV] Model vxrm in LLVM intrinsics and machine instructions for RVV fixed-point instructions](https://reviews.llvm.org/D151396) 
- D151397 [[3/3][RISCV] Model vxrm in C intrinsics for RVV fixed-point instruction vaadd](https://reviews.llvm.org/D151397) 
- D152879 [[RISCV] Model vxrm control for vsmul, vssra, vssrl, vnclip, and vnclipu](https://reviews.llvm.org/D152879)

Note that vsadd(u) and vssub(u) don't have the additional operand because computation of the instructions don't consider the rounding mode.


topperc:
Why not make the rounding mode a function argument?

eopXD:
> Why not make the rounding mode a function argument?

Yes this is also a choice. If as a function argument, with constraint of the value to be a constant, I think it is only a difference in style between the current proposal, which is statically encoding them in the function name.

I think my intuition comes from that the policy behaviors are also statically encoded in the function name.

topperc:
> > Why not make the rounding mode a function argument?
> 
> Yes this is also a choice. If as a function argument, with constraint of the value to be a constant, I think it is only a difference in style between the current proposal, which is statically encoding them in the function name.
> 
> I think my intuition comes from that the policy behaviors are also statically encoded in the function name.

The policy affects the number of function arguments in some cases so I think it had to be that way. That's not the case for rounding mode.

Making it an argument makes it easier to use as a macro argument or template parameter. Not sure if that's useful.

It also reduces the number of symbols, and I think makes the translation from C to LLVM IR intrinsics easier? We will want it as an argument in IR intrinsics.

nick-knight:
I agree, the the main reason we encoded "vtype" (output EEW, output EMUL, tail policy, mask policy) in the intrinsics names was because C doesn't support overloading (although some dialects do), and vtype affects the number and types of function arguments.

I think you are correct that `vxrm` doesn't affect the number and types of arguments, so we have more flexibility in this case. I prefer @eopXD's proposal for a superficial reason, that it feels to me to be more similar to the existing API. Plus, I'm not exactly sure how to constrain a C function argument to be a compile-time constant. 

Regarding using the C preprocessor to facilitate generic programming, we're already well accustomed to accommodating the existing API with [the usual tricks](https://github.com/flame/blis/blob/acd68f902cbf47299b280b7de8f1cc552ac0681b/kernels/sifive_x280/riscv_overloaded_intrinsics.h#L82-L83). Supporting `vxrm` could be achieved similarly. C++ wrappers could then expose `vxrm` via template non-type parameters, if desired. 

In any event, switching the fixed-point rounding mode tends to be an involved algorithmic decision, considered on an instruction-by-instruction basis. So I don't anticipate much demand for facilitating generic-`vxrm` programming. You typically do the math, determine the appropriate mode, then set it & forget it.

eopXD:
Although I personally don't consider the the number of intrinsics increased as a critical decision point between "static encode in function name" and "adding an extra parameter", I think I should still list them out here in detail so we can have full discussion.

As we want floating-point and fixed-point intrinsics have aligned naming scheme, and assuming 25% of instruction in RVV is floating-point instructions and 20% of instructions in RVV is fixed-point instructions, lets consider the increased amount of intrinsics. Note that we intend to keep the original floating-point intrinsics and remove the existing fixed-point intrinsics. That gives us:

- Static encode in function name results in `1/4 * 6 + 1/5 * 4 + (1 - 1/4 - 1/5) = 2.85`, which is **x2.85** of the current amount.
- Adding an extra parameter results in `1/4 * 2 + 1/5 + 1 - 1/4 - 1/5`, which is **x1.25** of the current amount.

kito-cheng:
I would support using parameter rather than static encode here, 2.85x is too huge number...

eopXD:
Change of interface as Kito is not a big fan of the growth of intrinsics.

Changing the addition from:

```cpp
- vint32m1_t __riscv_vsadd_vv_i32m1 (vint32m1_t op1, vint32m1_t op2, size_t vl);
+ vint32m1_t __riscv_vsadd_vv_i32m1_rnu (vint32m1_t op1, vint32m1_t op2, size_t vl);
+ vint32m1_t __riscv_vsadd_vv_i32m1_rne (vint32m1_t op1, vint32m1_t op2, size_t vl);
+ vint32m1_t __riscv_vsadd_vv_i32m1_rdn (vint32m1_t op1, vint32m1_t op2, size_t vl);
+ vint32m1_t __riscv_vsadd_vv_i32m1_rod (vint32m1_t op1, vint32m1_t op2, size_t vl);
```

to

```cpp
- vint32m1_t __riscv_vsadd_vv_i32m1 (vint32m1_t op1, vint32m1_t op2, size_t vl);
+ vint32m1_t __riscv_vsadd_vv_i32m1 (vint32m1_t op1, vint32m1_t op2, size_t vxrm, size_t vl);
```

PR description is updated.

topperc:
> Change of interface as Kito is not a big fan of the growth of intrinsics.
> 
> Changing the addition from:
> 
> ```c++
> - vint32m1_t __riscv_vsadd_vv_i32m1 (vint32m1_t op1, vint32m1_t op2, size_t vl);
> + vint32m1_t __riscv_vsadd_vv_i32m1_rnu (vint32m1_t op1, vint32m1_t op2, size_t vl);
> + vint32m1_t __riscv_vsadd_vv_i32m1_rne (vint32m1_t op1, vint32m1_t op2, size_t vl);
> + vint32m1_t __riscv_vsadd_vv_i32m1_rdn (vint32m1_t op1, vint32m1_t op2, size_t vl);
> + vint32m1_t __riscv_vsadd_vv_i32m1_rod (vint32m1_t op1, vint32m1_t op2, size_t vl);
> ```
> 
> to
> 
> ```c++
> - vint32m1_t __riscv_vsadd_vv_i32m1 (vint32m1_t op1, vint32m1_t op2, size_t vl);
> + vint32m1_t __riscv_vsadd_vv_i32m1 (vint32m1_t op1, vint32m1_t op2, size_t vxrm, size_t vl);
> ```
> 
> PR description is updated.

You can probably use `unsigned int` for the vxrm argument. It's required to be a constant so there shouldn't be any extend instructions generated for it being less than xlen.

zhongjuzhe:
It seems that you missed vssrl/vssra instructions ?

eopXD:
> It seems that you missed vssrl/vssra instructions ?
> 
> I saw RVV ISA SPEC:
> 
> These instructions shift the input value right, and round off the shifted out bits according to vxrm. The scaling right shifts have both zero-extending (vssrl) and sign-extending (vssra) forms. The data to be shifted is in the vector register group specified by vs2 and the shift amount value can come from a vector register group vs1, a scalar integer register rs1, or a zero-extended 5-bit immediate. Only the low lg2(SEW) bits of the shift-amount value are used to control the shift amount.
> 
> for vssrl/vssra

Nice catch. Fixed it in the latest commit.

zhongjuzhe:
Hi, could you update the enum in this PR today ?
https://gcc.gnu.org/pipermail/gcc-patches/2023-May/618827.html

auto-generated/api-testing/vaadd.c
```diff
@@ -5,354 +5,354 @@ typedef _Float16 float16_t;
 typedef float float32_t;
 typedef double float64_t;
 vint8mf8_t test_vaadd_vv_i8mf8(vint8mf8_t op1, vint8mf8_t op2, size_t vl) {
-  return __riscv_vaadd_vv_i8mf8(op1, op2, vl);
+  return __riscv_vaadd_vv_i8mf8(op1, op2, VXRM_RNU, vl);
```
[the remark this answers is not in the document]
eopXD: Adjusted the enum to

```c
enum __RISCV_VXRM {
  __RISCV_VXRM_RNU = 0,
  __RISCV_VXRM_RNE = 1,
  __RISCV_VXRM_RDN = 2,
  __RISCV_VXRM_ROD = 3,
};
```

topperc:
Other than them being considered "Fixed-point" instructions, is there any reason to do this for vsadd(u) and vssub(u)? The instructions don't use vxrm.

eopXD:
> Other than them being considered "Fixed-point" instructions, is there any reason to do this for vsadd(u) and vssub(u)? The instructions don't use vxrm.

Good point. I think them being "fixed-point instructions" are the only reason why they have a rounding mode operand. We will be introducing a corner case to the scheme if we don't add an operand for them.

topperc:
> > Other than them being considered "Fixed-point" instructions, is there any reason to do this for vsadd(u) and vssub(u)? The instructions don't use vxrm.
> 
> Good point. I think them being "fixed-point instructions" are the only reason why they have a rounding mode operand. We will be introducing a corner case to the scheme (the scheme of "fixed-point instructions have an additional operand for rounding mode control") if we don't add an operand for them.

Isn't the extra code generated by the compiler for an unnecessary vxrm write a bigger concern?

eopXD:
> > > Other than them being considered "Fixed-point" instructions, is there any reason to do this for vsadd(u) and vssub(u)? The instructions don't use vxrm.
> > 
> > 
> > Good point. I think them being "fixed-point instructions" are the only reason why they have a rounding mode operand. We will be introducing a corner case to the scheme (the scheme of "fixed-point instructions have an additional operand for rounding mode control") if we don't add an operand for them.
> 
> Isn't the extra code generated by the compiler for an unnecessary vxrm write a bigger concern?

You have a point here. We probably should remove the vxrm operand for vsadd(u)/vssub(u). This frees users from concerning the rounding mode when they are using the instruction. I will revise the PR and add a foot note for this under the rfc.md.

eopXD:
Remove the rounding mode operand in vsadd, vsaddu, vssub, and vssubu.

kito-cheng (approved):
Thanks :)

topperc:
Is __RISCV_VXRM_RNU and the other possible values documented?

eopXD:
> Is __RISCV_VXRM_RNU and the other possible values documented?

Yes, they are documented in [here](https://github.com/riscv-non-isa/rvv-intrinsic-doc/pull/222/commits/5f104245f0b6fe4e4afaf24b147c3d1befe7eca7)

rvv-intrinsic-rfc.md
```diff
@@ -464,6 +465,23 @@ vuint32m1_t vid_tamu(vbool32_t mask, vuint32m1_t merge, size_t vl);
 vuint32m1_t vid_v_u32m1_tama(vbool32_t mask, size_t vl);
 ```
 
+## Rounding mode in fixed-point intrinsics<a name="vxrm"></a>
+
+Regarding the fixed-point intrinsics, namely the intrinsics that represents instruction under [section 12 of the v-spec](https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc#sec-vector-fixed-point), the parameter that models the control of the rounding mode is identical to what is defined in [section 3.8 of the v-sepc](https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc#38-vector-fixed-point-rounding-mode-register-vxrm).
```
topperc, line 470: sepc -> spec

eopXD: Fixed, thank you for checking.

rvv-intrinsic-rfc.md
```diff
@@ -464,6 +465,23 @@ vuint32m1_t vid_tamu(vbool32_t mask, vuint32m1_t merge, size_t vl);
 vuint32m1_t vid_v_u32m1_tama(vbool32_t mask, size_t vl);
 ```
 
+## Rounding mode in fixed-point intrinsics<a name="vxrm"></a>
+
+Regarding the fixed-point intrinsics, namely the intrinsics that represents instruction under [section 12 of the v-spec](https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc#sec-vector-fixed-point), the parameter that models the control of the rounding mode is identical to what is defined in [section 3.8 of the v-spec](https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc#38-vector-fixed-point-rounding-mode-register-vxrm).
+
+Computation of `vsadd`, `vsaddu`, `vssub`, and `vssubu` does not need the rounding mode, therefore the intrinsics of these instructions does not have the parameter for rounding mode control.
```
topperc, line 472: "does not" -> "do not"

rvv-intrinsic-rfc.md
```diff
@@ -464,6 +465,23 @@ vuint32m1_t vid_tamu(vbool32_t mask, vuint32m1_t merge, size_t vl);
 vuint32m1_t vid_v_u32m1_tama(vbool32_t mask, size_t vl);
 ```
 
+## Rounding mode in fixed-point intrinsics<a name="vxrm"></a>
+
+Regarding the fixed-point intrinsics, namely the intrinsics that represents instruction under [section 12 of the v-spec](https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc#sec-vector-fixed-point), the parameter that models the control of the rounding mode is identical to what is defined in [section 3.8 of the v-spec](https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc#38-vector-fixed-point-rounding-mode-register-vxrm).
```
topperc, line 470: "represents instruction" -> "represent instructions"

eopXD:
@zixuan-wu You left this comment in the floating-point rounding mode PR, but I think this PR thread is a more appropriate place for me to reply to you.

> I am wondering if we need a dynamic mode enum which representing using current vxrm value? @eopXD

Unlike vector floating-point that has its existing scalar parts, the fixed-point instructions is an RVV-specific thing. So we have the freedom to not model a dynamic mode in the intrinsics.

The floating-point DOES define a default rounding mode, this gives dynamic mode (`DYN`) in floating-point a deterministic value when we assume users do not modify the rounding mode. So this gives us the contrary and pitfall if we try to model dynamic mode in fixed-point, which a default rounding mode is not defined.

On top of this, the C language does provide interface to rounding mode control (`#pragma STDC FENV_ACCESS`) for floating-point operations, but fixed-point doesn't, and if we decide to model dynamic mode in fixed-point, this would mean we have to come up with a similar `VXRM_ENV` and that may just complicate the use case when we know that most fixed-point users are advance users that DO care about the rounding mode, plus the RISC-V vector ABI has `vxrm` as call-clobbered too.

Concluding from the above rationales, we decide to not model a dynamic mode for the fixed-point intrinsics.

dzaima:
A slight thing of note is that `vnclip` with a shift of `0` can be used as a saturating narrow, but now requires a rounding mode parameter, even though it doesn't change any behavior. Compilers should optimize out configuring it ([compiler explorer](https://godbolt.org/z/zMK4ox1fb) - neither gcc nor clang do currently). (maybe it'd make sense to have intrinsics for saturating narrow by itself, but it's probably not significant enough to warrant that)

nick-knight:
Agreed. There are lots of cases where, for certain inputs, the rounding mode turns out to be irrelevant. (Although, most of the cases I can think of are less useful than the one you mention.) My opinion is that, for v1.0, we should make these optimizations a compiler problem. If we get significant pushback from intrinsics programmers, or the compiler problem turns out to be harder than expected, then we can always augment the API in the future.

Issue #223, opened 2023-04-26
Intrinsic Support for BF16 Extension


joshua-arch1:
The BF16 Extension has recently been proposed with three extra Instruction Set Extensions.
https://github.com/riscv/riscv-bfloat16

I'm wondering how we plan to address existing rvv intrinsics. Do we need to add new intrinsics tailored for bf16 datatype? If so, do we need to give a BF16 version for all the intrinsics with floating-point types? Maybe we can raise an issue for discussion.

joshua-arch1:
Maybe we can reuse most rvv intrinsics unless we want to generate new Zvfbfmin/Zvfbfwma instructions.

I have checked vector BF16 widening multiply-accumulate implementation in AArch64 sve. It uses llvm.aarch64.sve.fmlalb.nxv4f32 for fmlalb and llvm.aarch64.sve.bfmlalb for bfmlalb. Therefore what is certain for RISCV now is to define a new intrinsic for vfwmaccbf16.

That is to say, for llvm.riscv.vfwmacc.nxv8f32.nxv8f16, we need to use llvm.riscv.vfwmaccbf16 for a given bf16 format.

kito-cheng:
We definitely need to introduce new types like vbfloat16m1_t and corresponding RVV C intrinsic API.

> Maybe we can reuse most rvv intrinsics unless we want to generate new Zvfbfmin/Zvfbfwma instructions.
> 
> I have checked vector BF16 widening multiply-accumulate implementation in AArch64 sve. It uses > llvm.aarch64.sve.fmlalb.nxv4f32 for fmlalb and llvm.aarch64.sve.bfmlalb for bfmlalb. Therefore what is certain for RISCV now >is to define a new intrinsic for vfwmaccbf16.
>
>That is to say, for llvm.riscv.vfwmacc.nxv8f32.nxv8f16, we need to use llvm.riscv.vfwmaccbf16 in a given bf16 format.

That's LLVM implementation detail which should not specified in the intrinsic API spec.

joshua-arch1:
> We definitely need to introduce new types like vbfloat16m1_t and corresponding RVV C intrinsic API.
> 
> > Maybe we can reuse most rvv intrinsics unless we want to generate new Zvfbfmin/Zvfbfwma instructions.
[...]

PR #229, opened 2023-05-04
Update rvv-intrinsic-generator to define new RVV C intrinsic API for bf16 type

joshua-arch1:
As is duscussed in https://github.com/riscv-non-isa/rvv-intrinsic-doc/issues/223, the BF16 Extension has recently been proposed and we need to define new intrinsics for convert instruction (bf16-to-fp32/fp32-to-bf16), reinterpret function as well as vfwmaccbf16 in Zvfbfwma Extenstion.


circYuan:
Hi! It looks like the commits you provided are generating instructions like `vfncvtbf16_rtz_bf16_f_w_f16`. However, it seems like this instruction may not be correct because the spec doesn't define the `rtz` instruction for vector bf16 convert instructions. I think there may meed some condition checking on line 115, cvt_op_template.py.

eopXD (commented):
Thank you for pushing this and sorry I haven't been able to review this in the past 2 weeks.

Please separate the `vfwmaccbf16`, `vfwcvtbf16.f.f.v` and `vfncvtbf16.f.f.w` in a separate section. Please merge your commits in #224 to here too.

Regarding the type descriptions in the document, I think you should make it explicit that the bfloat16 types and the intrinsics will not be available when `zvbfmin` and `zvbfwma` is not specified in the architecture.

On the other hand, I agree with Rich's latest comment that it would be convenient to have a bfloat16 load/store that represents a (load + fncvt) and (fwcvt + store).

I have replied on my thought regarding this and the planned v1.0 release in the mailing list [0].

[0] https://lists.riscv.org/g/tech-rvv-intrinsics/message/57

joshua-arch1:
> On the other hand, I agree with Rich's latest comment that it would be convenient to have a bfloat16 load/store that represents a (load + fncvt) and (fwcvt + store).

I have updated my PR according to your comments, except the load/store intrinsics. Maybe you have a different understanding with Rich's. He meant an int16 load/store followed by a reinterpret cast rather than fncvt/fwcvt. Which definition is better?

eopXD:
Hi,

May you rebase upon the latest main? Thank you.

eopXD:
On top of rebasing, if possible, I think we can also have the effort of adding the intrinsics for Zvfbfwma and Zvfbfmin here. I imagine we would also need load/store intrinsics for the bfloat type. The narrow conversions and multiply-add instruction intrinsics will need a rounding mode variant.

kito-cheng:
I believe this should covered by this and https://github.com/riscv-non-isa/rvv-intrinsic-doc/pull/293 merged into main branch, so close this :)

PR #230, opened 2023-05-09
Add blank line before function instead of at the end

tclin914:


tclin914:
Does the extra line in the end-of-file have to exist?
This patch attempts to remove it.

eopXD:
I see my mistake, you are right.

Merged in f73ea2b4acc964a93ca5f7c97ce594f12cafb9d9 and cf21dcb3335541382e622e49ced7c0f8b78a35d3.

PR #233, opened 2023-06-01
Add missing `_mu` variants for some overloaded intrinsics

### LINKED ISSUE #232: Question about naming of overloaded policy load functions
Hi,
Why _mu series load intrinsics cannot be overloaded without type in the function name?

In [vector_loads_and_stores_functions](https://github.com/riscv-non-isa/rvv-intrinsic-doc/blob/master/auto-generated/policy_funcs/overloaded_intrinsic_funcs/01_vector_loads_and_stores_functions.md), __riscv_vle16_v_f16mf4_mu, __riscv_vle16_v_f16mf2_mu, etc., are with type in the function name.
But __riscv_vle16_tumu,__riscv_vle16_tum, etc., are without types in the function name.

Thanks.

eopXD:
Resolves #232


kito-cheng (approved):
LGTM, and @zhongjuzhe has implement that in GCC :)

PR #249, opened 2023-07-26
Remove the `vread_csr` / `vwrite_csr` intrinsics

### LINKED ISSUE #246: Will CSR API be removed?
I saw fixed-point and floating-point rounding API are merged.

But I saw:

Read/Write URW vector CSRs
enum RVV_CSR {
  RVV_VSTART = 0,
  RVV_VXSAT,
  RVV_VXRM,
  RVV_VCSR,
};

unsigned long __riscv_vread_csr(enum RVV_CSR csr);
void __riscv_vwrite_csr(enum RVV_CSR csr, unsigned long value);

unsigned long __riscv_vlenb();

Will they be removed?

eopXD:
These interfaces expose control to VSTART, VXSAT, VXRM, and VCSR. 

- The intrinsics should not expose control to VSTART as the intrinsic executions assume it to be set to 0.
- Rounding mode (`vxrm`) control is now modeled in the floating-point intrinsics.
- We are planning to add exception handling intrinsics variants in the future, that allows users to accss VXSAT for their fixed-point instruction executions.

Concluding these facts above, `vread_csr` and `vwrite_csr` is no longer needed.

---

Although we don't have the exception handling intrinsics for now, but for us to stay consistent that there won't be any backward incompatibility after v1.0, we should remove the interfaces now.

This resolves #246.

LLVM implementation is at https://reviews.llvm.org/D156321.

### DIFF
--- a/rvv-intrinsic-api.md
+++ b/rvv-intrinsic-api.md
@@ -61,16 +61,6 @@ These utility functions help users extract a smaller LMUL value from a larger LM
 ### Read/Write URW vector CSRs
 
 ```
-enum RVV_CSR {
-  RVV_VSTART = 0,
-  RVV_VXSAT,
-  RVV_VXRM,
-  RVV_VCSR,
-};
-
-unsigned long __riscv_vread_csr(enum RVV_CSR csr);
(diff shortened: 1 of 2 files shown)


kito-cheng (approved):
LGTM

Issue #252, opened 2023-08-02
Examples use arithmetic on `void*`, which isn't allowed by the c standard


camel-cdr:
Examples, specifically rvv_memcpy,  use arithmetic on `void*`, which isn't allowed by the c standard, see: https://stackoverflow.com/questions/3523145/pointer-arithmetic-for-void-pointer-in-c#3524270

nick-knight:
Feel free to submit a PR. These examples are not a normative part of the spec, so the task group is treating them as lower priority, to focus our limited (volunteer) resources on reaching API freeze/ratification.

camel-cdr:
👍 I'll look into it when I've got the time

Issue #244, opened 2023-07-13
strcmp example is wrong


dzaima:
[`examples/rvv_strcmp.c`](https://github.com/riscv-non-isa/rvv-intrinsic-doc/blob/eebf706e58288a741dc99b3a05120fbeddbf0de6/examples/rvv_strcmp.c#L6-L25) incorrectly manages vl - the vl of the second load (`vl1`) goes completely unused, and thus the function could do all kinds of broken behavior if `src1`'s loaded element count is greater than that of `src2`.

A fix could be to make the second load take & update the `vl` of the first load, matching the behavior of the assembly reference.

PR #254, opened 2023-08-09
Fix and cleanup examples

camel-cdr:
This resolves https://github.com/riscv-non-isa/rvv-intrinsic-doc/issues/252, https://github.com/riscv-non-isa/rvv-intrinsic-doc/issues/244,  unsigned char instead of char, and cleans up the code a bit.

The problem with using `char` in combination with `vint8mX_t` is that `char` can be either signed or unsigned, so to remedy this problem, I've changed the examples to use `unsigned char` explicitly (and `vuint8m8_t`).

[rvv_reduce.c](https://github.com/riscv-non-isa/rvv-intrinsic-doc/commit/ac6d76f822764cef333ed5dd7c4e8d6798e7cb80) had an error, where a agnostic tail policy could make it produce the wrong results, this has been fixed by using the `_tu` intrinsic variant.

### DIFF
--- a/examples/rvv_reduce.c
+++ b/examples/rvv_reduce.c
@@ -7,7 +7,7 @@ void reduce_golden(double *a, double *b, double *result_sum,
   int count = 0;
   double s = 0;
   for (int i = 0; i < n; ++i) {
-    if (a[i] != 0.0) {
+    if (a[i] != 42.0) {
       s += a[i] * b[i];
       count++;
     }
@@ -17,8 +17,8 @@ void reduce_golden(double *a, double *b, double *result_sum,
   *result_count = count;
 }
 
-void reduce(double *a, double *b, double *result_sum, int *result_count,
-              int n) {
+void reduce_vec(double *a, double *b, double *result_sum, int *result_count,
+                int n) {
   int count = 0;
   // set vlmax and initialize variables
   size_t vlmax = __riscv_vsetvlmax_e64m1();
@@ -30,9 +30,9 @@ void reduce(double *a, double *b, double *result_sum, int *result_count,
     vfloat64m1_t vec_a = __riscv_vle64_v_f64m1(a, vl);
     vfloat64m1_t vec_b = __riscv_vle64_v_f64m1(b, vl);
 
-    vbool64_t mask = __riscv_vmfne_vv_f64m1_b64(vec_a, vec_zero, vl);
+    vbool64_t mask = __riscv_vmfne_vf_f64m1_b64(vec_a, 42, vl);
 
-    vec_s = __riscv_vfmacc_vv_f64m1_m(mask, vec_s, vec_a, vec_b, vl);
+    vec_s = __riscv_vfmacc_vv_f64m1_tumu(mask, vec_s, vec_a, vec_b, vl);
     count = count + __riscv_vcpop_m_b64(mask, vl);
   }
   vfloat64m1_t vec_sum;
@@ -57,7 +57,7 @@ int main() {
   double golden_sum, actual_sum;
   int golden_count, actual_count;
   reduce_golden(A, B, &golden_sum, &golden_count, N);
-  reduce(A, B, &actual_sum, &actual_count, N);
+  reduce_vec(A, B, &actual_sum, &actual_count, N);
 
   // compare
   puts(golden_sum - actual_sum < 1e-6 && golden_count == actual_count ? "pass"
--- a/examples/rvv_strcmp.c
+++ b/examples/rvv_strcmp.c
@@ -3,21 +3,20 @@
 #include <string.h>
 
 // reference https://github.com/riscv/riscv-v-spec/blob/master/example/strcmp.s
-int strcmp_vec(const char *src1, const char *src2) {
+int strcmp_vec(const char *source1, const char *source2) {
+  const unsigned char *src1 = (const void*)source1, *src2 = (const void*)source2;
   size_t vlmax = __riscv_vsetvlmax_e8m2();
   long first_set_bit = -1;
-  size_t vl, vl1;
-  while (first_set_bit < 0) {
-    vint8m2_t vec_src1 = __riscv_vle8ff_v_i8m2(src1, &vl, vlmax);
-    vint8m2_t vec_src2 = __riscv_vle8ff_v_i8m2(src2, &vl1, vlmax);
+  size_t vl;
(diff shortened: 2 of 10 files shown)


examples/rvv_reduce.c
```diff
@@ -30,9 +30,9 @@ void reduce(double *a, double *b, double *result_sum, int *result_count,
     vfloat64m1_t vec_a = __riscv_vle64_v_f64m1(a, vl);
     vfloat64m1_t vec_b = __riscv_vle64_v_f64m1(b, vl);
 
-    vbool64_t mask = __riscv_vmfne_vv_f64m1_b64(vec_a, vec_zero, vl);
+    vbool64_t mask = __riscv_vmfne_vf_f64m1_b64(vec_a, 0, vl);
 
-    vec_s = __riscv_vfmacc_vv_f64m1_m(mask, vec_s, vec_a, vec_b, vl);
+    vec_s = __riscv_vfmacc_vv_f64m1_tu(vec_s, vec_a, vec_b, vl);
```
eopXD, line 35: Does this change the behavior?

examples/rvv_strcmp.c
```diff
@@ -4,24 +4,23 @@
...
 
     first_set_bit = __riscv_vfirst_m_b4(vec_terminate, vl);
-    src1 += vl;
-    src2 += vl;
   }
-  src1 -= vl - first_set_bit;
-  src2 -= vl - first_set_bit;
-  return *src1 - *src2;
+  s1 -= vl - first_set_bit;
+  s2 -= vl - first_set_bit;
```
eopXD, line 22: Does this fix the problem mentioned in #244?

eopXD (commented):
Thank you for creating this pull request! The simplification looks clean.

The only objection I have is that I don't find abbreviating `src`/`dst` to `s`/`d` very informative.

eopXD:
Nit: May you add "Resolve #XXX" into the commit messages so they can be interlinked with the issue?

examples/rvv_reduce.c
camel-cdr: Yes, but in a good way:

If we want to use a vector register to collect all sums and reduce afterward, we need to make sure the tail policy is set to undisturbed, otherwise the last iteration may overwrite the inactive elements.
The mask is redundant, because multiplying by zero results in zero.

examples/rvv_strcmp.c
camel-cdr: Changing `__riscv_vle8ff_v_i8m2(src2, &vl1, vlmax)` to `__riscv_vle8ff_v_u8m2(s2, &vl, vl);`  on line 13 should have fixed it.

examples/rvv_reduce.c
eopXD: I see your point. My intent was to demonstrate the masked instructions though, lets change the if statement to checking something that is not zero?

eopXD: Let's change the name to `reduce_vec` too?

examples/rvv_strcmp.c
eopXD: I see the fix now. Thank you for explaining.

camel-cdr:
> The only objection I have is that I don't find abbreviating src/dst to s/d very informative.

I haven't addressed this yet, what name would you suggest?

I've also wondered about the choices of LMUL, was the idea to show a few different ones? Because usually maximizing LMUL given the number of required registers seems to be the best option.

eopXD:
> > The only objection I have is that I don't find abbreviating src/dst to s/d very informative.
> 
> I haven't addressed this yet, what name would you suggest?
> 

Welp, maybe `input_src`/`input_dst` for parameter and `src` / `dst` for iteration?

> I've also wondered about the choices of LMUL, was the idea to show a few different ones? Because usually maximizing LMUL given the number of required registers seems to be the best option.

Yes the intent was to show different LMUL's.

camel-cdr:
Ok, I've now used `source`, `destination`, `src`, `dst`.

examples/rvv_reduce.c
```diff
@@ -30,9 +30,9 @@ void reduce(double *a, double *b, double *result_sum, int *result_count,
     vfloat64m1_t vec_a = __riscv_vle64_v_f64m1(a, vl);
     vfloat64m1_t vec_b = __riscv_vle64_v_f64m1(b, vl);
 
-    vbool64_t mask = __riscv_vmfne_vv_f64m1_b64(vec_a, vec_zero, vl);
+    vbool64_t mask = __riscv_vmfne_vf_f64m1_b64(vec_a, 42, vl);
 
-    vec_s = __riscv_vfmacc_vv_f64m1_m(mask, vec_s, vec_a, vec_b, vl);
+    vec_s = __riscv_vfmacc_vv_f64m1_tum(mask, vec_s, vec_a, vec_b, vl);
```
eopXD, line 35: Supposedly we shouldn't need the tail elements because we are doing reduction for the given `n` elements right?

camel-cdr: No, the reduction happens outside of the loop, so the last iteration could cause elements we want to reduce later to b overwritten. Come to think of it, this should be `_tumu`.

dzaima: yeah, `_tumu` is needed. (as more clarification on `_tu` - this intrinsic will update the same `vec_s` with both `vl==VLMAX` and `vl<VLMAX`, and we want to keep the top `VLMAX-vl` items when transitioning from an iteration of the former to the latter,)

camel-cdr: qemu and spike should imho default to writing all ones to inactive elements in the agnostic policy, or at least add the option to enable this. It's way too easy to write and successfully test broken code. (I may look into this, but I haven't worked with the codebase before)

dzaima: It seems QEMU does support this: `-cpu rv64,v=on,vext_spec=v1.0,vlen=256,rvv_ta_all_1s=on,rvv_ma_all_1s=on`

Neat!

camel-cdr: Looks like @eopXD actually [added this](https://lists.nongnu.org/archive/html/qemu-riscv/2022-06/msg00050.html), great!
I only saw the code implementing it for load/stores when looking through the code, but this make rvv_reduce fail with just the `_m` variant.

eopXD: My question is resolved. Thank you guys!

eopXD (approved):

PR #264, opened 2023-08-30
Supported gnu test

xl0714:
Hi, @eopXD 
This PR is what I did together with @Incarnation-p-lee  and @yulong-plct  to generate gnu gcc test cases.
1. Currently we put the generated test cases in the directory "gnu-auto-generated", do you have any suggestions?
2. We have prepared a gnu gcc test cases CI: "gnu-gcc-daily-build.yml", all gnu gcc test cases are currently passed.


kito-cheng:
Could you reorg/squash commits into few small pieces? it's not easy to review for such # of commit and changes.

kito-cheng:
Also plz rebase instead of introduce merge commit

PR #267, opened 2023-09-08
Remove overloaded variants of masked vector load instruction intrinsics (seg and non-seg)

eopXD:
The intrinsics contain all scalar parameters and by the limitation of
the C language we are not able to implement overloaded variants of them
in the compiler.

This was an oversight in the specification. The LLVM test suite passes
the clang front-end test case because the test case only checks if LLVM
IR can be generated. In fact, the test cases can not be compiled successfully.


dzaima:
While unmasked loads cannot be overloaded, masked ones technically can (and, it seems, are handled correctly by both gcc and clang - https://godbolt.org/z/Gbe3GaqP5), due to the `mask`/`maskedoff` argument type effectively including LMUL information. Granted, this is pretty weird, and without `maskedoff` it relies on the pointer element type, and the clang error message for the unmasked case seems to be slightly broken. (I don't have much of an opinion on whether these should be kept or removed)

eopXD:
@dzaima Thank you for checking. I intended to remove those mis-printed ones and not the ones we already support, which are the overloaded variants of the unmasked load instructions. I will update the patch.

eopXD:
Hmmm, after compiling with RISC-V headers I realized that this PR is totally redundant. We do have the supported prototypes printed, and un-supported ones omitted. Sorry for the noise.

PR #271, opened 2023-09-14
Polishing to the specification

eopXD:


doc/rvv-intrinsic-spec.adoc
```diff
@@ -464,19 +555,42 @@ The intrinsics provide extra utility functions to help users manipulate across t
...
+
+For intrinsics with no `vd` operand, the compiler is free to pick any register as the destination register.
+
+=== Assumption of `vstart=0` for intrinsics users.
+
+As the v-spec has mentioned:
+
+> The vstart CSR is writable by unprivileged code, but non-zero vstart values may cause vector instructions to run substantially slower on some implementations, so vstart should not be used by application programmers.
+
+The RVV intrinsics are designed for application users. The intrinsics will perform the semantic of vector instructions under `vstart=0`.
```
nick-knight, line 575: Personally, I think portability (not performance) is a stronger rationale to hide `vstart` from the intrinsics API.

```suggestion

The `vstart` CSR is currently not exposed to the intrinsics programmer, 
and the intrinsics have the semantics of `vstart = 0`.
Support for positive `vstart` values is implementation defined,
thus portable application software should not set `vstart > 0`.
```

nick-knight (changes requested):
Thanks @eopXD for all your improvements to the intrinsics documentation.

I haven't read through it all, I'm just responding to your change regarding `vstart`, since this came up in a separate conversation (about why the intrinsics API allows for undisturbed elements in `vslideup` but not `vslidedown`).

doc/rvv-intrinsic-spec.adoc
eopXD: Good point. Updated the note, thank you.

doc/rvv-intrinsic-spec.adoc
```diff
@@ -96,29 +98,33 @@ enum __RISCV_VXRM {
...
+NOTE: The implicit-`frm` intrinsics are intended to be used regardless of `FENV_ACCESS`. They are provided when `FENV_ACCESS` is on for the (few) programmers who are already using fenv; and they are provided when `FENV_ACCESS` is off for the (vast majority of) programmers who prefer the default rounding mode.
 
-[[chapter-explicit-frm]]
+[[explicit-frm]]
 ==== Explicit-`frm` intrinsics
 
 The explicit-`frm` intrinsics contain the `frm` operand which indicates the rounding mode (`frm`) ^10^ control. The floating-point intrinsics with the `frm` operand are followed by an `_rm` suffix in the function name.
 
-The `frm` operand is required to be an integer constant expression. The compiler should implement the following enum that maps to the defined rounding mode values under RISC-V ISA Manual Table 8.1 ^8^.
+The `frm` operand is required to be an integer constant expression. The compiler should implement the following enum that maps to the defined rounding mode values under RISC-V ISA Manual Table 8.1 ^9^.
```
dzaima, line 127: there was a "integer constant expression" → "constant integer expression" change in another place, should it be here too?

doc/rvv-intrinsic-spec.adoc
```diff
@@ -96,29 +98,33 @@ enum __RISCV_VXRM {
...
 
 ==== Implicit-`frm` intrinsics
 
 The implicit-`frm` intrinsics behave like any C-language floating-point expressions, using the default rounding mode when `FENV_ACCESS` is off, and using the `fenv` dynamic rounding mode when `FENV_ACCESS` is on.
 
 NOTE: Both GNU and LLVM compilers generate scalar floating-point instructions using dynamic rounding mode, relying on the kernel initialization to set `frm` to `RNE` (specified as "roundTiesToEven" in IEEE-754 (a.k.a. IEC 60559)).
 
-NOTE: The implicit-`frm` intrinsics are intended to be used regardless of `FENV_ACCESS`. They are provided when `FENV_ACCESS` is on for the (few) programmers who are already using fenv. And they are provided when `FENV_ACCESS` is off for the (vast majority of) programmers who prefer the default rounding mode.
+NOTE: The implicit-`frm` intrinsics are intended to be used regardless of `FENV_ACCESS`. They are provided when `FENV_ACCESS` is on for the (few) programmers who are already using fenv; and they are provided when `FENV_ACCESS` is off for the (vast majority of) programmers who prefer the default rounding mode.
```
dzaima, line 120: "intended to be used" → "intended to be usable" maybe?

doc/rvv-intrinsic-spec.adoc
```diff
@@ -268,16 +270,20 @@ vbool4_t __riscv_vreinterpret_v_i32m1_b4 (vint32m1_t src);
 [[implicit-naming-scheme]]
 === Implicit (Overloaded) naming scheme
 
-The overloaded interface aims to let users put values of different EEW and EMUL as the input operand, therefore hiding the EEW and EMUL encoded in the function name. The `_rm` prefix for explicit-`frm` intrinsics (<<chapter-control-to-frm>>) is also hidden. The intrinsics under this scheme are the "overloaded intrinsics", which in parallel we have "non-overloaded intrinsics" defined under <<explicit-naming-scheme>>.
+The implicit (overloaded) interface aims to provide a generic interfaces taking values of different EEW and EMUL as the input. Therefore the implicit intrinsics hide the EEW and EMUL encoded in the function name. The `_rm` prefix for explicit-`frm` intrinsics (<<control-of-frm>>) is also hidden. The intrinsics under this scheme are the "overloaded intrinsics", which in parallel we have the "non-overloaded intrinsics" defined under <<explicit-naming-scheme>>.
```
dzaima, line 273: "generic interfaces" → "generic interface"

doc/rvv-intrinsic-spec.adoc
```diff
@@ -28,64 +28,66 @@ To leverage the intrinsics in the compiler, the header `<riscv_vector.h>` needs
...
 
 === Control of `vl`
 
-Direct control of `vl` (vector length register) ^11^  is not exposed at the C language level. In the intrinsics that represents vector instructions, the `vl` operand indicates the active vector length valid during intrinsic execution.
+Direct control of `vl` (vector length register) ^11^  is not exposed at the C language level. The RVV C intrinsics provide intrinsics with a `vl` operand to indicate the active vector length valid during intrinsic execution. The compiler is responsible to set the correct value to `vl` given the informed active vector length.
 
 NOTE: Intrinsics for instructions that behave the same with different `vl` settings (e.g. `vmv.s.x`) do not have a `vl` operand.
 
-NOTE: Intrinsics will at most operate `vlmax` (derived in 3.4.2 of v-spec) elements.
+NOTE: Intrinsics will at most operate on `VLMAX` ^5^ elements.
```
dzaima, line 68: maybe should explicitly describe the semantics for when a `vl` argument is greater than `VLMAX`? (i.e. whether the compiler may treat it as undefined behavior, or if there's specific behavior (presumably equivalent to it being passed through a respective `__riscv_vsetvl_*`?))

doc/rvv-intrinsic-spec.adoc
```diff
@@ -431,52 +431,195 @@ The mask types encode the ratio that is derived from `EEW`/`EMUL`. The mask type
...
 == Pseudo intrinsics
 
 The intrinsics provide extra utility functions to help users manipulate across the RVV intrinsic types. These functions are called "pseudo intrinsics". These pseudo intrinsics do not represent any real instructions.
 
-[[pseudo-vsetvl-vsetvlmax]]
-=== `vsetvl`/`vsetvlmax`
+[[pseudo-vsetvl]]
+=== `vsetvl`
+
+The `vsetvl` intrinsics return `min(avl, VLMAX)` given the provided element width and LMUL in the suffix. Please refer to how VLMAX is defined in the v-spec ^5^.
```
dzaima, line 544: `min(avl, VLMAX)` doesn't match `vset{i}vl{i}` in the V spec, where for `VLMAX < avx < 2*VLMAX` it can be anything in `[ceil(VLMAX/2), VLMAX]`

doc/rvv-intrinsic-spec.adoc
```diff
@@ -431,52 +431,195 @@ The mask types encode the ratio that is derived from `EEW`/`EMUL`. The mask type
...
+The `vreinterpret` intrinsics are provided for users to go across the strongly-typed scheme. The intrinsic is limited to conversion between types operating upon the same number of registers.
+
 [[pseudo-vundefined]]
 === `vundefined`
+The `vundefined` intrinsics are placeholders for `vset` and `vcreate` to represent agnostic values.
+
+[[pseudo-vget]]
+=== `vget`
+
+The `vget` intrinsics allow users to get smaller LMUL values from larger LMUL ones .The `vget` intrinsics also allows users to extract non-tuple (`NFIELD=1`) types from tuple (`NFIELD>1`) types after segment load intrinsics.
```
dzaima, line 563: "ones .The" → "ones. The"

doc/rvv-intrinsic-spec.adoc
```diff
@@ -1,24 +1,24 @@
 == Introduction
 
-The RISC-V vector C intrinsics provides users interfaces in the C language level to directly leverage the RISC-V "V" extension ^0^ (also abbreviated as "RVV"), with help from the compiler to handle instruction scheduling and register allocation.
+The RISC-V vector C intrinsics provides users interfaces in the C language level to directly leverage the RISC-V "V" extension ^0^ (also abbreviated as "RVV"), with help from the compiler to handle instruction scheduling and register allocation. The RVV C intrinsics also aims to free users from responsibility to maintain the correct configuration setting ^18^ for the vector instruction executions.
 
-This document will use the term "v-spec" to indicate the RISC-V "V" extension specification ^0^.
+This document uses the term "v-spec" to indicate the RISC-V "V" extension specification ^0^. This document also uses the term "RVV" as an abbreviation for the RISC-V "V" extension.
```
nick-knight, line 5: > This document uses the term "v-spec" to indicate the RISC-V "V" extension specification ^0^

We wouldn't say "f-spec" to refer to the F-extension specification. These are ratified standard extensions of the RISC-V ISA. I think we should just say "the ISA spec" (or "specification").

Many standard extensions are developed in repositories separate from https://github.com/riscv/riscv-isa-manual , but after ratification they are merged into this document. Here is the PR to merge the (ratified) vector extension: https://github.com/riscv/riscv-isa-manual/pull/1088.

Until this is merged, it is OK to cite https://github.com/riscv/riscv-v-spec. Or perhaps cite both, with a comment in the bibliography.

nick-knight, line 5: > This document uses the term "v-spec" to indicate the RISC-V "V" extension specification ^0^

We wouldn't say "f-spec" to refer to the F-extension specification. These are ratified standard extensions of the RISC-V ISA. I think we should just say "the ISA spec" (or "specification").

Many standard extensions are developed in repositories separate from https://github.com/riscv/riscv-isa-manual , but after ratification they are merged into this document. Here is the PR to merge the (ratified) vector extension: https://github.com/riscv/riscv-isa-manual/pull/1088.

Until this is merged, it is OK to cite https://github.com/riscv/riscv-v-spec. Or perhaps cite both, with a comment in the bibliography.

nick-knight (changes requested):

doc/rvv-intrinsic-spec.adoc
eopXD: Adjusted wording to use "the specification" instead of v-spec.

doc/rvv-intrinsic-spec.adoc
eopXD: Added description to clarify the semantics. Please check again, thank you.

doc/rvv-intrinsic-spec.adoc
eopXD: Added description to clarify the semantics. Please check again, thank you.

dzaima:
Is the intent of a `__riscv_vsetvl_*` intrinsic be exactly `min(avl, VLMAX)`, and not be possible to map directly to a `vsetvli rd, rs1, ...` instruction without some additional computation of the `min()`?

If that's the case, then the current LLVM and GCC implementations are non-conformative, as they do map directly to `vsetvli`: https://godbolt.org/z/bzYG3vxzM.

For example, on a VLEN=128 implementation (i.e. `VLMAX=4` for e32m1), a `wrap(6)` call would result in a `vsetvli` instruction that is allowed to, on a RVV1.0-compliant RISC-V implementation, produce a `vl` of either `3` or `4`, as per "`ceil(AVL / 2) ≤ vl ≤ VLMAX` if `AVL < (2 * VLMAX)`" of [section 6.3 of the V spec](https://dzaima.github.io/intrinsics-viewer/data/v-spec.html#_constraints_on_setting_vl), and thus `wrap(6)` could equal `3`, while this document specifies `min(avl,VLMAX) = min(6,4) = 4`.

eopXD:
> Do I understand correctly that the intent of a `__riscv_vsetvl_*` intrinsic is to be exactly `min(avl, VLMAX)`, and thus be impossible to map directly to a `vsetvli rd, rs1, ...` instruction without some additional computation of the `min()`?
> 
> If that's the case, then the current LLVM and GCC implementations are non-conforming, as they do map directly to `vsetvli`: https://godbolt.org/z/bzYG3vxzM.
> 
> For example, on a VLEN=128 implementation (i.e. `VLMAX=4` for e32m1), a `wrap(6)` call of the LLVM/GCC output would result in a `vsetvli` instruction that is allowed to, on a RVV1.0-compliant RISC-V implementation, produce a `vl` of either `3` or `4`, as per "`ceil(AVL / 2) ≤ vl ≤ VLMAX` if `AVL < (2 * VLMAX)`" of [section 6.3 of the V spec](https://dzaima.github.io/intrinsics-viewer/data/v-spec.html#_constraints_on_setting_vl), and thus `wrap(6)` could equal `3`, while this document specifies `min(avl,VLMAX) = min(6,4) = 4`.

Ah, yes. You have found the hidden piece of the intrinsic implementation. However if you see this snippet https://godbolt.org/z/Wx4TPq7fs, the compiler is not respecting the element width and length multiplier provided by `vsetvl`. The vsetvl intrinsic is not intended to map to an exact instruction, the compiler is responsible to insert the appropriate vsetvl instruction.

topperc:
> > Do I understand correctly that the intent of a `__riscv_vsetvl_*` intrinsic is to be exactly `min(avl, VLMAX)`, and thus be impossible to map directly to a `vsetvli rd, rs1, ...` instruction without some additional computation of the `min()`?
> > If that's the case, then the current LLVM and GCC implementations are non-conforming, as they do map directly to `vsetvli`: https://godbolt.org/z/bzYG3vxzM.
> > For example, on a VLEN=128 implementation (i.e. `VLMAX=4` for e32m1), a `wrap(6)` call of the LLVM/GCC output would result in a `vsetvli` instruction that is allowed to, on a RVV1.0-compliant RISC-V implementation, produce a `vl` of either `3` or `4`, as per "`ceil(AVL / 2) ≤ vl ≤ VLMAX` if `AVL < (2 * VLMAX)`" of [section 6.3 of the V spec](https://dzaima.github.io/intrinsics-viewer/data/v-spec.html#_constraints_on_setting_vl), and thus `wrap(6)` could equal `3`, while this document specifies `min(avl,VLMAX) = min(6,4) = 4`.
> 
> Ah, yes. You have found the hidden piece of the intrinsic implementation. However if you see this snippet https://godbolt.org/z/Wx4TPq7fs, the compiler is not respecting the element width and length multiplier provided by `vsetvl`. The vsetvl intrinsic is not intended to map to an exact instruction, the compiler is responsible to insert the appropriate vsetvl instruction.

The compiler is still required respect the VLMAX implied by the SEW and LMUL. This means the compiler must respect the ratio between SEW and LMUL given to the intrinsic. The intrinsic must return the value the hardware vsetvli intruction would return for that VLMAX which is not always min(avl, VLMAX).

dzaima:
Right, I am not saying that `__riscv_vsetvl_e32m1` should necessarily map to `vsetvli rd, rs1, e32, m1`; indeed, `vsetvli rd, rs1, e32, m1` and `vsetvli rd, rs1, e64, m2` will result in the same `rd` value, and thus a compiler can choose to use those interchangeably for calculating `rd`.

What I am saying is that `__riscv_vsetvl_e32m1(avl)`, if specified to be `min(avl, VLMAX)`, _can not be implemented as_ `vsetvli rd, x[avl], e32, m1` (or `..., e64, m1`, etc), even if the compiler wanted to do so, because the behavior of that `vsetvli` instruction is not guaranteed to be `min(avl, VLMAX)`, as per the aforementioned section 6.3. of the V extension 1.0 specification.

doc/rvv-intrinsic-spec.adoc
```diff
@@ -1,165 +1,180 @@
...
+[[control-of-vl]]
 === Control of `vl`
 
-Direct control of `vl` (vector length register) ^11^  is not exposed at the C language level. In the intrinsics that represents vector instructions, the `vl` operand indicates the active vector length valid during intrinsic execution.
+Direct control of `vl` (vector length register) ^11^  is not exposed at the C language level. The RVV C intrinsics provide intrinsics with a `vl` operand to indicate the active vector length valid during intrinsic execution. The compiler is responsible to set the correct value to `vl` given the informed active vector length.
 
 NOTE: Intrinsics for instructions that behave the same with different `vl` settings (e.g. `vmv.s.x`) do not have a `vl` operand.
 
-NOTE: Intrinsics will at most operate `vlmax` (derived in 3.4.2 of v-spec) elements.
+NOTE: The intrinsics follow the specification ^27^ and will at most operate on `VLMAX` ^5^ elements. If a value larger than `VLMAX` is provided to `vl`, the intrinsics will operate on `VLMAX` elements. Essentially the number of elements processed will match what is returned when provided to `__riscv_vsetvl_*` <<pseudo-vsetvl>>.
```
nick-knight, line 72: This is not exactly correct, as discussed by @dzaima and @topperc in the PR conversation. In this case, I think it is better to say less. Here is a proposed fix.

```suggestion
The actual value written to the `vl` control register is subject to implementation defined behavior and is typically not known until runtime.
The intrinsics programmer should use the `__riscv_vsetvl_*` intrinsics <<pseudo-vsetvl>> to determine its value,
for use (e.g.) in loop stripmining logic.
```

nick-knight (changes requested):
Some suggestions related to the `vl` issues.

doc/rvv-intrinsic-spec.adoc
nick-knight: On second thought, this might warrant further discussion. For known VLEN or sufficiently small AVL, the programmer might want to avoid stripmining and rely on the fact that `vl` = AVL when AVL < VLMAX. The way I've worded this might allow an implementation to pick `vl` < AVL in this case.

nick-knight: I just made an edit to my proposed fix. I'm writing this message so you see a notification about it.

eopXD:
@nick-knight @dzaima @topperc Just made an update to description associated with "control of `vl`" and pseudo intrinsics `vsetvl` and `vsetvlmax`. Please check again, thank you.

eopXD:
Ping.

eopXD:
Ping.

doc/rvv-intrinsic-spec.adoc
```diff
@@ -28,64 +28,66 @@ To leverage the intrinsics in the compiler, the header `<riscv_vector.h>` needs
...
+For example, RVV type `vint64m1_t` and `__riscv_vle64_v_i64m1` is not available under architecture `rv64gc_zve32x`.
 
-[[control-to-vector-programming-model]]
+[[control-of-vector-programming-model]]
 == Control of the vector extension programming model
 
-The intrinsics allow users to control fields in `vtype`, and also rounding modes for fixed-point (`vxrm`) and floating-point intrinsics (`frm`).
+The intrinsics allow users to control fields in `vtype`, and also rounding modes for fixed-point (`vxrm`) and floating-point (`frm`) vector computations.
 
 In this section, we cover how the intrinsics embed the control of `vtype` fields in the function names. Please see <<naming-scheme>> for the rules described in this section.
```
rofirrim: "In this chapter" perhaps?

doc/rvv-intrinsic-spec.adoc
```diff
@@ -96,29 +98,33 @@ enum __RISCV_VXRM {
...
-For the floating-point intrinsics, representing the instructions under v-spec section 13, the intrinsics have two classes, called implicit-`frm` and explicit-`frm`.
+For the floating-point intrinsics, representing the floating-point arithmetic instructions ^23^, the intrinsics have two variants, called the implicit-`frm` and the explicit-`frm` intrinsics.
+
+NOTE: Control to the floating-point accrued exceptions flag fields (`fflag`) ^10^ is not yet covered in the RVV C intrinsics v1.0. We plan to support them in follow-up versions in a compatible way with existing intrinsics in v1.0.
 
 ==== Implicit-`frm` intrinsics
 
 The implicit-`frm` intrinsics behave like any C-language floating-point expressions, using the default rounding mode when `FENV_ACCESS` is off, and using the `fenv` dynamic rounding mode when `FENV_ACCESS` is on.
 
 NOTE: Both GNU and LLVM compilers generate scalar floating-point instructions using dynamic rounding mode, relying on the kernel initialization to set `frm` to `RNE` (specified as "roundTiesToEven" in IEEE-754 (a.k.a. IEC 60559)).
```
rofirrim: I assume this "kernel" is the OS kernel but maybe this can be expressed more generically as the "environment initialization" because this might be actually done by the loader of the C library.

doc/rvv-intrinsic-spec.adoc
```diff
@@ -180,45 +188,48 @@ vint32m1_t __riscv_vadd_vv_i32m1_tumu(vbool32_t mask, vint32m1_t maskedoff,
...
 
 In general, the intrinsics are encoded as the following. The intrinsics under this naming scheme are the "non-overloaded intrinsics", which in parallel we have the "overloaded intrinsics" defined under <<implicit-naming-scheme>>.
 
+The naming rules is the following.
+
 ```
 __riscv_{V_INSTRUCTION_MNEMONIC}_{OPERAND_MNEMONIC}_{RETURN_TYPE}_{ROUND_MODE}_{POLICY}{(...)
 ```
 
 * `OPERAND_MNEMONIC` are like `vv`, `vx`, `vs`, `vvm`, `vxm`
```
rofirrim: I understand `vle` and `vse` also include the case `v`?

doc/rvv-intrinsic-spec.adoc
```diff
@@ -180,45 +188,48 @@ vint32m1_t __riscv_vadd_vv_i32m1_tumu(vbool32_t mask, vint32m1_t maskedoff,
...
-The general naming scheme is not sufficient to express intrinsics. The exceptions are enumerated under <<explicit-exception-naming>>.
+The general naming scheme is not sufficient to express intrinsics. The exceptions are enumerated in the proceeding section <<explicit-exception-naming>>.
 
 [[explicit-exception-naming]]
 === Exceptions in the explicit (non-overloaded) naming scheme
 
-This section enumerates the exceptions in the naming scheme.
+This section enumerates the exceptions in the explicit (non-overloaded) naming scheme.
 
 ==== Scalar move instructions
```
rofirrim: I think for consistency the listings here should use `___riscv__vmv_x_s_...` instead of `vmv_x_s_...`.

Same for `vredsum_vs_...` below.

doc/rvv-intrinsic-spec.adoc
```diff
@@ -268,16 +279,20 @@ vbool4_t __riscv_vreinterpret_v_i32m1_b4 (vint32m1_t src);
 [[implicit-naming-scheme]]
 === Implicit (Overloaded) naming scheme
 
-The overloaded interface aims to let users put values of different EEW and EMUL as the input operand, therefore hiding the EEW and EMUL encoded in the function name. The `_rm` prefix for explicit-`frm` intrinsics (<<chapter-control-to-frm>>) is also hidden. The intrinsics under this scheme are the "overloaded intrinsics", which in parallel we have "non-overloaded intrinsics" defined under <<explicit-naming-scheme>>.
+The implicit (overloaded) interface aims to provide a generic interfaces taking values of different EEW and EMUL as the input. Therefore the implicit intrinsics hide the EEW and EMUL encoded in the function name. The `_rm` prefix for explicit-`frm` intrinsics (<<control-of-frm>>) is also hidden. The intrinsics under this scheme are the "overloaded intrinsics", which in parallel we have the "non-overloaded intrinsics" defined under <<explicit-naming-scheme>>.
```
rofirrim: This is purely style, so feel free to leave the text as is: I'd replace the uses of hide/hidden with omit/omitted.

doc/rvv-intrinsic-spec.adoc
```diff
@@ -316,202 +351,362 @@ The following intrinsics have an irregular naming pattern.
...
-| Types     | EMUL=1/8      | EMUL=1/4       | EMUL=1/ 2      | EMUL=1        | EMUL=2        | EMUL=4        | EMUL=8
-| int8_t    | *vint8mf8_t*  | vint8mf4_t     | vint8mf2_t     | vint8m1_t     | vint8m2_t     | vint8m4_t     | vint8m8_t
-| int16_t   | N/A           | *vint16mf4_t*  | vint16mf2_t    | vint16m1_t    | vint16m2_t    | vint16m4_t    | vint16m16_t
-| int32_t   | N/A           | N/A            | *vint32mf2_t*  | vint32m1_t    | vint32m2_t    | vint32m4_t    | vint32m32_t
-| int64_t   | N/A           | N/A            | N/A            | *vint64m1_t*  | *vint64m2_t*  | *vint64m4_t*  | *vint64m8_t*
-| uint8_t   | *vuint8mf8_t* | vuint8mf4_t    | vuint8mf2_t    | vuint8m1_t    | vuint8m2_t    | vuint8m4_t    | vuint8m8_t
-| uint16_t  | N/A           | *vuint16mf4_t* | vuint16mf2_t   | vuint16m1_t   | vuint16m2_t   | vuint16m4_t   | vuint16m8_t
-| uint32_t  | N/A           | N/A            | *vuint32mf2_t* | vuint32m1_t   | vuint32m2_t   | vuint32m4_t   | vuint32m8_t
-| uint64_t  | N/A           | N/A            | N/A            | *vuint64m1_t* | *vuint64m2_t* | *vuint64m4_t* | *vuint64m8_t*
+| Types | EMUL=1/8 | EMUL=1/4 | EMUL=1/ 2 | EMUL=1 | EMUL=2 | EMUL=4 | EMUL=8
```
rofirrim: In the table, I wonder if rather than using bold face for the types unavailable under Zve32* we could use a small marker such as ᵃ (e.g. maybe using a superscript `^a^`) after the type and then explain after the table that those types are not available in Zve32.

Bold face makes those type stand out too much but maybe this is not easy to achieve in AsciiDoc. In such case feel free to ignore my suggestion :)

doc/rvv-intrinsic-spec.adoc
```diff
@@ -316,202 +351,362 @@ The following intrinsics have an irregular naming pattern.
...
-[[pseudo-vlmul_trunc-vlmul_ext]]
-=== `vlmul_trunc`/`vlmul_ext`
+The `vset` intrinsics allow users to combine small LMUL values into larger LMUL ones. The `vset` intrinsics also allows users to combine non-tuple (`NFIELD=1`) types to tuple (`NFIELD>1`) types for segment store intrinsics.
+
+The intrinsics do not map to any real instruction. Whether if the implementation will generate vector move instructions is an optimization issue.
+
+[[pseudo-vlmul_trunc]]
+=== `vlmul_trunc`
+
+The `vlmul_trunc` intrinsics are syntactic sugars that essentially have the same semantic as `vget` with `idx=0`.
```
rofirrim: This is style probably: the "essentially" may suggest that there is some slight difference (even if small). If the semantics is exactly the same due to the syntactic sugar nature of those intrinsics, I'd drop the word.

doc/rvv-intrinsic-spec.adoc
```diff
@@ -316,202 +351,362 @@ The following intrinsics have an irregular naming pattern.
...
+
+In other words, the compiler does not guarantee generating the instruction for all memory accesses in strided load/store intrinsics with a stride of 0. If the user needs all memory accesses to be performed, they should use an indexed load/store intrinsics with all zero indices.
+
+=== Leveraging instructions with operand mnemonics of `vi`
 
-In other words, compiler does not guarantee generating the all memory accesses instruction in strided load/store intrinsics with stride of 0. If the user needs all memory accesses to be performed, they should use an indexed load/store intrinsics with all zero indices.
+The intrinsics provide variants with operand mnemonics of `vv` and `vx`, but not `vi`. This was an intentional design to reduce the total amount of out-going intrinsics. The implementation is responsible to emit instructions with operand mnemonics of `vi` when an immediate that can be expressed within 5-bit is provided to the intrinsics.
```
rofirrim: This sounds like a mandatory thing. Like if an implementation had to honour this. Is this the intent?

If this wasn't the intent, maybe rephrasing the last statement to something like "It is a quality of implementation issue whether instructions with `vi` operand mnemonics are emitted when an immediate operand can be encoded in the 5-bit field of the a `vi` instruction".

rofirrim (commented):
Thanks @eopXD polishing the document.

I left some comments, none of them too critical.

doc/rvv-intrinsic-spec.adoc
eopXD: Yes, I think the intent is to not have a `vi` variant and whether or not the compiler is able to recognize 5-bit immediate and use the `vi` instruction is a quality issue.

nick-knight (approved):
While I think there is room for further wordsmithing, I think the technical content of the documentation is sufficient to present this to the Software HC.

eopXD:
Merging this PR now. Thanks all people that helped to work this through!

PR #272, opened 2023-09-15
Supported gnu auto generated test cases

yulong-plct:
Hi, @eopXD :
This PR adds the implementation for gnu auto generated test case that I did together with @Incarnation-p-lee and @xl0714.
Please review it.
It is worth noting that this PR only generates cases, and the rest of the content will be in another PR.
Thanks!

PR #275, opened 2023-09-18
Add gnu auto generated test cases to the generator

yulong-plct:
Hi,@eopXD:
This PR add the codes that can generate gnu test cases to the generator.
I did it together with @Incarnation-p-lee and @xl0714.
After that, I would also submit PR about creating folders, adding headers and checker.
Please review it. Thanks!


eopXD (commented):
The CI currently does not pass for this single commit. I think a whole series of commit just like what you did in #272 would be good, with the comments addressed and commits reorganized.

yulong-plct:
OK.I'll complete the other commits later.

yulong-plct:
I added the other two commits about creating folders and adding headers.  We consider to create anther PR for adding checker, because we need to fully consider how to better interpret the magic number in the check information.@eopXD @Incarnation-p-lee @xl0714

rvv-intrinsic-generator/Makefile
```diff
@@ -108,7 +108,7 @@ endef
 define gen_tests
 	rm -f $(1)/*.c
 	$(PY3) -m $(MAIN) --gen $(2) \
-		--out $(1) $(3) \
+		--out $(1) $(3) $(4)\
```
eopXD, line 111: No extra parameter is needed in this function. Your patch runs successfully without this edit.

rvv-intrinsic-generator/rvv_intrinsic_gen/main.py
```diff
@@ -96,7 +98,7 @@ class GenTypes:
           "overloaded-docs", "non-overloaded-test", "overloaded-test",
           "non-overloaded-compatible-header", "overloaded-compatible-header"
       ])
-  parser.add_argument("--llvm", default=False, action="store_true")
+  parser.add_argument("--toolchain-type", default=ToolChainType.UNKNOWN)
```
eopXD, line 101: Need code to stop the generator if no `--toolchain-type` is specified.

eopXD (commented):
The PR should be good after the remaining comments are addressed.

rvv-intrinsic-generator/rvv_intrinsic_gen/main.py
yulong-plct: I tried  add the code in the generator.APITestGenerator.Funs function, but will affect the auto generated test cases at the auto-generated/API folder.  If we make sure that the toolchain-type specified in the Makefile is correct, can we not add the corresponding code?

yulong-plct: I tried  add the code in the generator.APITestGenerator.Funs function, but will affect the auto generated test cases at the auto-generated/API folder.  If we make sure that the toolchain-type specified in the Makefile is correct, can we not add the corresponding code?

eopXD: I mean, handle error under main.py to exit and emit error when `--toolchain-type` is not specified.

yulong-plct: OK! I got it. I will add code.

eopXD: Thanks for the reminder, we can still generate test case, like what we have right now in `api-testing/`, that don't need LLVM or GNU test headers. So I think no modifications is needed here.

eopXD (approved):

PR #283, opened 2023-10-08
Add a workflow to test latest GNU GCC toolchain compilation for the intrinsics

yulong-plct:
Hi, @eopXD:
This PR contains two commits. The first is to modify the checker to ensure that all gnu auto generated cases were passed. The second is to add a workflow that triggers the CI tests.
Please review, thanks a lot!
Of course, co-developers are still @xl0714 and @Incarnation-p-lee


eopXD (approved):
Looks neat! Thank you.

PR #287, opened 2023-10-24
Add more details in the specification

### LINKED ISSUE #31: Use fixed-width integer type instead of long?
In a number of places, the V-extension intrinsics API uses `long` or `unsigned long` to represent an XLEN-bit integer (signed or unsigned, resp.). It's true that `long` has size XLEN in the currently defined ILP32 and LP64 ABIs
https://github.com/riscv/riscv-elf-psabi-doc/blob/master/riscv-elf.md#-named-abis
However, note that

    A future version of this specification may define an ILP32 ABI for the RV64 ISA

And one can imagine a similar scenario for RV128. Hence, I think it's preferable to future-proof the intrinsics API by using typedefs [Footnote 1]. I don't think this adds much notational overhead.

Please note that we had the same discussion for the P-extension intrinsics, and arrived at the same conclusion
https://lists.riscv.org/g/tech-p-ext/topic/73148653#26

[Footnote 1]
On GCC and Clang/LLVM, you can alias the relevant types as follows:
```
#include <inttypes.h>
#if __riscv_xlen == 32
typedef int32_t int_xlen_t;
typedef uint32_t uint_xlen_t;
#elif __riscv_xlen == 64
typedef int64_t int_xlen_t;
typedef uint64_t uint_xlen_t;
#endif
```
Those who enjoy preprocessor wizardry might prefer something like
```
#define create_type2(a, b, c) a ## b ## c
#define create_type1(a, b, c) create_type2(a, b, c)
typedef create_type1(int,__riscv_xlen,_t) int_xlen_t;
typedef create_type1(uint,__riscv_xlen,_t) uint_xlen_t;
```

### LINKED ISSUE #74: Utility of vmmv.m and vmv.v.v intrinsics
Shouldn't users be encouraged to use operator= to copy vectors like any other variable? Using a mv intrinsic guarantees an instruction is emitted by the compiler even if register allocation would allow the input and output to use the same register. Unless we teach the compiler to look for these operations and turn them into copies that register allocation can coalesce.

### LINKED ISSUE #166: Should vcpop.m return type be size_t?
Current interface
```
unsigned long vcpop_m_b1 (vbool1_t op1, size_t vl);
```

However, the semantic of vcpop.m is
>  The `vcpop.m` instruction counts the number of mask elements of the active elements of the vector source mask register that have the value 1 and writes the result to a scalar `x` register.

That means the maximum value depends on VL. And VL related interface is
```
size_t vsetvl_e8mf8 (size_t avl);
size_t vsetvlmax_e8mf8 ();
```

So, should we change the interface to
```
size_t vcpop_m_b1 (vbool1_t op1, size_t vl);
```

### LINKED ISSUE #242: Document relationship between inline asm and the intrinsics
With the intrinsics modeling rounding mode control (and the compiler handling vsetvl insertions), we should let users be aware of this and keep their hands away from `vtype` and the rounding mode CSR-s (`frm`, `vxrm`) in their inline assembly.

This issue is a placeholder to keep this in mind in later document polishing phase.

eopXD:
While cleaning up the issue threads, I found some more details we need to have in the specification.

This PR:
- Resolves #74 by adding a description of "Semantic of assignment (`=`) in the C language" under the programming note section of the spec
- Resolves #242 by adding a description of "Mixing inline assembly and intrinsics" under the programming note section of the spec
- Replace `float16_t` with `_Float16` that is now supported by both LLVM and GCC.  This keeps us away from the need to define our own type. (Resolves #31, and alongside also resolves #166.
- Adds missing description to the `vlenb` utility function
- Adds description to `new_vl` argument in fault-only-first load intrinsics

### DIFF
--- a/doc/rvv-intrinsic-spec.adoc
+++ b/doc/rvv-intrinsic-spec.adoc
@@ -621,14 +621,20 @@ The `vlenb` intrinsic returns what is held inside the read-only CSR `vlenb` ^29^
 
 [,c]
 ----
-unsigned vlenb();
+unsigned __riscv_vlenb();
 ----
 
 == Programming Notes
 
-=== Semantic of assignment (`=`) in the C language
+=== Copying vector register group contents
 
-Without optimizations, the compiler is conservative and assumes an object of RVV type to represent an allocated space on the stack. The assignment operator (`=`) represents a whole register store for the object on the right hand side, and a load for the object on the left hand side.
+There is no intrinsic that directly maps to the whole vector register move instructions (`vmv<nr>r.v`) ^30^.
+
+For copying of the vector contents in whole, we encourage the users to use the assignment operator (`=`).
+
+The assignment operator (`=`) represents the semantic of a whole vector register (group) copy for the expression on the right hand side to the RVV type object on the left hand side. The semantic will still maintain a whole vector register content copy for fractional LMUL types. This enables the compiler to coalesce register usage when possible.
+
+Users may leverage the vector move intrinsics (`vmv_v_v`) intrinsics if they hope to copy vector register groups with `vl != VLMAX`.
 
 === The passthrough (`vd`) argument in the intrinsics
 
@@ -671,7 +677,7 @@ The compiler will be conservative to registers (`vtype`, `vxrm`, `frm`) when enc
 
 === The `new_vl` argument in fault-only-first load intrinsics
 
-The fault-only-first load intrinsics write the new value inside the `vl` register into the address of the `new_vl` argument. Providing an illegal memory location will result in a memory access violation.
+The fault-only-first load intrinsics write the new value inside the `vl` register into the address of the `new_vl` argument. Providing an illegal memory location is undefined behavior.
 
 == References
 
@@ -736,3 +742,5 @@ NOTE: Standard extensions are merged into `riscv/riscv-isa-manual` after ratific
 ^28^Section 6.4 (Example of stripmining and changes to SEW) in the specification ^0^
 
 ^29^Section 3.6 (Vector Byte Length `vlenb`) in the specification ^0^
+
+^30^Section 16.6 (Whole Vector Register Move) in the specification ^0^
--- a/doc/rvv-intrinsic-spec.adoc
+++ b/doc/rvv-intrinsic-spec.adoc
@@ -409,7 +409,7 @@ Due to the limitation of the C language (without the aid of features like C++ te
 
 The intrinsics are designed to be strongly-typed. The intrinsics provide `vreinterpret` intrinsics to help users go across the strongly-typed scheme if necessary.
 
-Non-mask (integer and floating-point) data types have SEW and LMUL encoded. 
+Non-mask (integer and floating-point) data types have SEW and LMUL encoded.
 
 [[integer-type]]
 === Integer types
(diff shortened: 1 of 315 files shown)


rofirrim (approved):
LGTM. Thanks @eopXD !

doc/rvv-intrinsic-spec.adoc
```diff
@@ -614,8 +614,22 @@ The `vlmul_ext` intrinsics are syntactic sugars that have the same semantic as `
...
 The `vcreate` intrinsics are syntactic sugars for tuple types creation. They have the same semantic as multiple `vset`-s filling in values accordingly into the tuple type.
 
+[[pseudo-vlenb]]
+=== `vlenb`
+
+The `vlenb` intrinsic returns what is held inside the read-only CSR `vlenb` ^29^, which is the vector register length in bytes.
+
+[,c]
+----
+unsigned vlenb();
```
dzaima, line 624: shouldn't this be `unsigned __riscv_vlenb();`?

doc/rvv-intrinsic-spec.adoc
```diff
@@ -651,6 +665,14 @@ The intrinsics provide variants with operand mnemonics of `vv` and `vx`, but not
 
 It is an optimization issue for the implementation to emit instructions with operand mnemonics of `vi` when an immediate that can be expressed within 5-bit is provided to the intrinsics.
 
+=== Mixing inline assembly and intrinsics
+
+The compiler will be conservative to registers (`vtype`, `vxrm`, `frm`) when encountering inline assembly. Users should be aware that mixing uses of intrinsics and inline assembly will result in extra save and restore.
+
+=== The `new_vl` argument in fault-only-first load intrinsics
+
+The fault-only-first load intrinsics write the new value inside the `vl` register into the address of the `new_vl` argument. Providing an illegal memory location will result in a memory access violation.
```
dzaima, line 674: Probably should say that an illegal memory location will result in undefined behavior? Otherwise this differs from the way regular C/C++ stores behave. Both clang and gcc already treat a `new_vl == NULL` as undefined behavior and don't necessarily generate a memory access violation.

doc/rvv-intrinsic-spec.adoc
```diff
@@ -614,8 +614,22 @@ The `vlmul_ext` intrinsics are syntactic sugars that have the same semantic as `
...
+[,c]
+----
+unsigned vlenb();
+----
+
 == Programming Notes
 
+=== Semantic of assignment (`=`) in the C language
+
+Without optimizations, the compiler is conservative and assumes an object of RVV type to represent an allocated space on the stack. The assignment operator (`=`) represents a whole register store for the object on the right hand side, and a load for the object on the left hand side.
```
topperc, line 631: The load on the right hand side isn't quite correct. Consider

```
a = vadd(b, c)
```

There are 2 loads on the right side for b and c, and they both occur before the vadd. Not because of the `=`.

eopXD: Improved the description by discussing what to do when users want to copy vector register group contents, which is what we really wanted to clarify.

topperc (approved):
LGTM

dzaima (approved):

nick-knight:
@topperc recently noticed that the `vcpop` and `vfirst` intrinsics had been switched from `unsigned long` and `long` to `unsigned int` and `int`. This appears to have happened in this PR.

This PR description says:
> (Resolves https://github.com/riscv-non-isa/rvv-intrinsic-doc/issues/31, and alongside also resolves https://github.com/riscv-non-isa/rvv-intrinsic-doc/issues/166.

This was not the agreed-upon resolution to those issues.

PR #292, opened 2023-11-04
Fixes for the specification

eopXD:
- Leveling of the printed list of intrinsic prototypes was wrong
- Use `_Float16`/`float`/`double` for scalar types of floating-point types with width of 16/32/64. Although we have `<stdfloat.h>` [0] introduced in C++23, this is not fully supported in the upstream compilers (LLVM/GCC). This is fixed by leveraging the `TypeHelper` class instead of deriving locally in the templates.

These changes will be cherry-picked to `v1.0.x` branch.

[0] https://en.cppreference.com/w/cpp/types/floating-point


eopXD:
@Amanieu Please checkout the added notes to uninitialized/`vundefined` values in (https://github.com/riscv-non-isa/rvv-intrinsic-doc/pull/292/commits/68f125b818d5570f7bbc81182e6c413f1bda1d88).

doc/rvv-intrinsic-spec.adoc
```diff
@@ -624,6 +624,12 @@ unsigned __riscv_vlenb();
 
 == Programming Notes
 
+=== Uninitialized variables with RVV types and `vundefined`
+
+Object created in the declaration of an RVV type represents a scalable-sized object that corresponds to a memory address. The default values inside such uninitialized values, and also to variables being assigned to the `vundefined` intrinsics represent non-deterministic values. Users should be aware to not rely on the result of these values.
+
+For example, a `vxor` operation to two either uninitialized or `vundefined` call will not yield a zero.
```
rofirrim, line 631: An object is missing here, I think you meant "uninitialized variables".

I suggest phrasing this as "For example, a `vxor` operation whose operands are either two uninitialized variables or two function calls to `vundefined` may not yield a zero" (will not yield a zero might be too strong).

rofirrim (commented):
All changes make sense to me. Just a minor comment.

Thanks @eopXD

eopXD:
Rebased and added commit [c33f99f](https://github.com/riscv-non-isa/rvv-intrinsic-doc/pull/292/commits/c33f99fb054417d45176280708642a324e8d14c4) that addresses Rogers comments.

doc/rvv-intrinsic-spec.adoc
```diff
@@ -626,9 +626,9 @@ unsigned __riscv_vlenb();
 
 === Uninitialized variables with RVV types and `vundefined`
 
-Object created in the declaration of an RVV type represents a scalable-sized object that corresponds to a memory address. The default values inside such uninitialized values, and also to variables being assigned to the `vundefined` intrinsics represent non-deterministic values. Users should be aware to not rely on the result of these values.
+Variables created in the declaration of an RVV type represents a scalable-sized object that corresponds to a memory address. The default values inside such uninitialized variables, and also to variables being assigned to the `vundefined` intrinsics, represent non-deterministic values. Users should be aware to not rely on the result of these values.
 
-For example, a `vxor` operation to two either uninitialized or `vundefined` call will not yield a zero.
+For example, a `vxor` operation whose operands are either two uninitialized variables or two function calls to `vundefined` may not yield a zero.
```
Amanieu, line 631: In other words, what you are saying here is that performing an arithmetic operation using uninitialized values is valid, and produces a new uninitialized value.

I would prefer a stricter definition which says that behavior is undefined when operating on uninitialized elements: you must ensure that any uninitialized elements are disabled by vl or masking so that no arithmetic is performed using them.

doc/rvv-intrinsic-spec.adoc
```diff
@@ -626,9 +626,9 @@ unsigned __riscv_vlenb();
 
 === Uninitialized variables with RVV types and `vundefined`
 
-Object created in the declaration of an RVV type represents a scalable-sized object that corresponds to a memory address. The default values inside such uninitialized values, and also to variables being assigned to the `vundefined` intrinsics represent non-deterministic values. Users should be aware to not rely on the result of these values.
+Variables created in the declaration of an RVV type represents a scalable-sized object that corresponds to a memory address. The default values inside such uninitialized variables, and also to variables being assigned to the `vundefined` intrinsics, represent non-deterministic values. Users should be aware to not rely on the result of these values.
```
Amanieu, line 629: * It's not clear what "corresponds to a memory address" means here.
* It should be clarified that this also applies to masked-out elements (with "ma") and tail elements (with "ta"): these also become uninitialized values.

topperc:
I'm not sure I understand why vundefined intrinsics exist. I thought some of their original usage was due to some intrinsic signatures that took passthru operands before we defined explicit policy intrinsics. And we needed to use vundefined to get tail agnostic on those intrinsics. Hopefully that's been resolved with the current interface? If not we should fix the interface.

Some conversations I've had seem to suggest that vundefined could be used to suppress -Wunitialized by explicitly stating that the variable is undefined. But I don't think that was the original intent and I don't know if I would advocate for that usage.

nick-knight:
@topperc I agree, Craig. One of my goals with the policy intrinsics was to eliminate the `vundefined` intrinsics altogether. But in the confusion of reintroducing tuple types, when we forgot to reintroduce `vcreate`, the `vundefined` intrinsics became briefly useful again as a way of suppressing `-Wuninitialized`. I'd certainly support killing them off once and for all.

dzaima:
For the case of tuples or building a register group, `vcreate` does indeed pretty much replace `vundefined`+`vget`+`vset`, except of course the case where you do want undefined elements passed to the following operation. Though I don't know how often, if at all, that'd be a real need (something like, `vmerge`ing a bunch of things together in a loop, and starting the loop with a `vundefined`? but you can just peel back the first iteration. A reasonably-realistic thing might be doing a `vrgather` where the two vector arguments don't have the same LMUL (which they don't at all need to for reasonable functionality; of course it'd be better if RVV1.0 allowed specifying the LMUL the operands separately for this, but ¯\\\_(ツ)\_/¯)).

But, for what it's worth, you can get a semantically-identical result to e.g. `__riscv_vundefined_u32m1()` with `__riscv_vid_v_u32m1(0)` (it's shorter even!), so undefined values stay regardless of if they're via a clear `vundefined` or not.

eopXD:
@topperc @nick-knight @dzaima 
I think that "the redundancy of `vundefined`" should belong to a separate thread so we can converge this pull request. The note on "agnostic values" covers `vundefined` because it exists now in the specification. If any further decision is made the note will be adjusted.

@Amanieu Just updated the description. I realize its not about the stack or memory address the spec should be described, but the definition of what are the agnostic (indeterminate) values we are dealing with in the RVV C intrinsics.

Amanieu (approved):
New wording LGTM.

kito-cheng (approved):
LGTM

PR #304, opened 2023-12-22
Remove a redundant space in LLVM float testcases header

tclin914:
It has two spaces between `-disable-O0-optnone` and `\`.


kito-cheng (approved):
Thanks for refine this :)

PR #306, opened 2023-12-26
Remove suffix `_v` from vlmul_trunc and vlmul_ext to be the same with others.

tclin914:


kito-cheng:
Sorry, no incompatible change allowed at this moment, we are at the last mile of the stable release.

PR #307, opened 2023-12-26
Directly add prefix `v` into op name for vreinterpret, vsetvl and vsetvlmax intrinsics

tclin914:


kito-cheng (approved):
Thanks, it more consistent with other misc stuffs.

tclin914:
Add vsetvl and vsetvlmax cases.

Issue #322, opened 2024-04-29
Clarify what float and double means


rofirrim:
This is a comment from the ARC review.

We should clarify, in accordance to the RISC-V ABI what `float `and `double` means, for their precise meaning is platform dependent in the C standard.

kito-cheng:
Fixed via #324

PR #324, opened 2024-04-29
Clarify what we mean by float and double

rofirrim:
Address #322

### DIFF
--- a/doc/rvv-intrinsic-spec.adoc
+++ b/doc/rvv-intrinsic-spec.adoc
@@ -431,6 +431,8 @@ Types with subscript "^\*^" are available when `ELEN >= 64` (that is, unavailabl
 [[floating-point-type]]
 === Floating-point types
 
+NOTE: This specification uses `_Float16` to designate IEEE-754 binary16, `float` to designate IEEE-754 binary32 and `double` to designate IEEE-754 binary64.
+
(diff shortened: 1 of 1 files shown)


doc/rvv-intrinsic-spec.adoc
```diff
@@ -431,6 +431,8 @@ Types with subscript "^\*^" are available when `ELEN >= 64` (that is, unavailabl
 [[floating-point-type]]
 === Floating-point types
 
+NOTE: This specification uses `_Float16` to designate IEEE-754 binary16, `float` to designate IEEE-754 binary32 and `double` to designate IEEE-754 binary64.
```
4vtomat, line 434: Do we also need bfloat16?

nick-knight: BF16 was not part of the original vector extensions. We will add it at a later date, once we get the base vector intrinsics ratified. Similarly for the vector crypto stuff. (Arguably, we could drop the FP16 stuff too, since Zvfh wasn't part of the original extension.)

4vtomat: Got it, thanks @nick-knight !

kito-cheng (approved):
LGTM, thanks :)

PR #328, opened 2024-05-06
Add a quick-build target to Makefile

kito-cheng:
Full PDF generation will spend lot of time due to compile all of the function list, however most time we edit the document we don't really care about that part, so I try to add a quick-build target to reduce the PDF build time for quick preview for the changes.


rofirrim (approved):
Haha, I did a similar thing in my branch.

Let's land this for better QoL during development.

Thanks @kito-cheng !

PR #330, opened 2024-05-07
Clarify the current support of `vxsat`

### LINKED ISSUE #318: Clarify the consequences of vxsat not being handled by the intrinsics
As pointed by the ARC review it is acceptable that we leave out the vxsat support for a 1.0 version of the specification but we should clarify the consequences. The spec already contains a reference to this but it seems reasonable to expand on the consequences.

In summary: The specification does not provide a mechanism to read `vxsat` or update it in a predictable way using a sequence of fixed point arithmetic intrinsics.

rofirrim:
Fixes #318


rofirrim:
Thanks @nick-knight !

PR #333, opened 2024-05-15
Add a section with examples

### LINKED ISSUE #319: Add a section with examples
The ARC review suggests that we can improve readability by adding some examples.

My suggestion is make intrinsic versions of the examples in the V extension specification (found in https://github.com/riscv/riscv-v-spec/tree/master/example ).

rofirrim:
The examples are non-normative. I've taken a subset of the examples in the examples directory of the repository.

This fixes #319

### DIFF
--- a/doc/rvv-intrinsic-examples.adoc
+++ b/doc/rvv-intrinsic-examples.adoc
@@ -1,0 +1,297 @@
+== Examples
+
+NOTE: This section is non-normative.
+
+NOTE: No claims about efficiency are made the examples presented in this section.
+
+This section presents examples that use the RVV intrinsics specified in this
+document. The examples are in C and assume `#include <riscv_vector.h>` has
+appeared earlier in the source code.
+
+=== Memory copy
+
+.An implementation of the `memcpy` function of the C Standard library using RVV intrinsics.
+====
+[,c]
+----
+void *memcpy_rvv(void *restrict destination, const void *restrict source,
+    size_t n) {
+  unsigned char *dst = destination;
+  const unsigned char *src = source;
+  // copy data byte by byte
+  for (size_t vl; n > 0; n -= vl, src += vl, dst += vl) {
+    vl = __riscv_vsetvl_e8m8(n);
+    // Load src[0..vl)
+    vuint8m8_t vec_src = __riscv_vle8_v_u8m8(src, vl);
+    // Store dst[0..vl)
+    __riscv_vse8_v_u8m8(dst, vec_src, vl);
+    // src is incremented vl (bytes)
+    // dst is incremented vl (bytes)
+    // n is decremented vl
+  }
+  return destination;
+}
+----
+====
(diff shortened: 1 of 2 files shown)


doc/rvv-intrinsic-examples.adoc
```diff
@@ -0,0 +1,297 @@
...
+        vl = __riscv_vsetvl_e64m1(k);
+
+        // Load row a[i][k..k+vl)
+        vfloat64m1_t vec_a = __riscv_vle64_v_f64m1(ptr_a, vl);
+        // Load column b[k..k+vl)[j]
+        vfloat64m1_t vec_b =
+          __riscv_vlse64_v_f64m1(ptr_b, sizeof(double) * m, vl);
+
+        // Accumulate dot product of row and column. If vl < vlmax
+        // we need to preserve the existin values of vec_s, hence the
```
topperc: existing*

rofirrim:
@nick-knight can you take a closer look at the matrix multiplication? I took the existing example but it presumes a transposed matrix in B (I think) so I reimplemented it for a more naive approach so we can show a strided access. Also I understand partial accumulations should use tail undisturbed (regardless of the fact that this is a vfmacc with already 3 input operands).

doc/rvv-intrinsic-examples.adoc
```diff
@@ -0,0 +1,297 @@
...
+    // elements of a that are not 42.
+    vbool64_t mask = __riscv_vmfne_vf_f64m1_b64(vec_a, 42, vl);
+
+    // for all e in [0..vl)
+    //  vec_s[e] ← vec_s[e] + vec_a[e] * vec_b[e], if mask[e] is enabled
+    //             vec_s[e]                      , otherwise (mask undisturbed)
+    vec_s = __riscv_vfmacc_vv_f64m1_tumu(mask, vec_s, vec_a, vec_b, vl);
+
+    // Adds to count the number of elements in mask that are enabled.
+    count = count + __riscv_vcpop_m_b64(mask, vl);
```
topperc, line 283: could use `+=` here

nick-knight:
@rofirrim I've always strongly disliked the matmul example in this repo: it's not how a sane person would implement it.

camel-cdr:
I agree with @nick-knight, I think we should stick with code we can confidently say will portable perform at close to the peak performance.

kito-cheng (approved):
LGTM, and...

> I think the examples should be that, examples, and not necessarily a reference or library of efficient functions. The matrix multiply example is intentionally qualifed as "naive" in the examples for this reason.

+1 for that, the main goal of those example in this document is demonstrate the intrinsic programming model rather than providing best implementation :)

rofirrim:
Hi @kito-cheng thanks a lot for merging this for me.

I will update `v1.0.x` to the current `main` so not to stall further work (such as bf16).

PR #336, opened 2024-06-05
[NFC] Make Generator class becomes ABC

jerryzj:
- The following member functions will return NotImplemented if not implemented in derived classes to let user aware that the function call is not functioning
  - write()
  - write_title()
  - gen_prologue()
  - inst_group_prologue()
  - inst_group_epilogue()
  - post_gen()

- The func function is set as abstract method, all the derived classes shoule have their own implementation

- The original func implementation is copied to DocGenerator and APITestGenerator and replace the call of base func() implementation


kito-cheng (approved):
LGTM, waiting CI run :)

PR #343, opened 2024-06-10
Change return type of vcpop and vfirst back to `unsigned long` and `long` respectively

### LINKED ISSUE #342: `vfirst` and `vcpop` return types unexpectedly changed
(Cross-posting https://github.com/riscv-non-isa/rvv-intrinsic-doc/pull/287#issuecomment-2159273686 to a new issue.)

In https://github.com/riscv-non-isa/rvv-intrinsic-doc/pull/287, the return values for the `vcpop` and `vfirst` intrinsics were switched from `unsigned long` and `long` to `unsigned int` and `int`, while the implementations in upstream LLVM and GCC still have the former types. This API change appears to have been made in this PR, with the following justification:

> Replace `float16_t` with `_Float16` that is now supported by both LLVM and GCC. This keeps us away from the need to define our own type. (Resolves https://github.com/riscv-non-isa/rvv-intrinsic-doc/issues/31, and alongside also resolves https://github.com/riscv-non-isa/rvv-intrinsic-doc/issues/166.

FP16 is not relevant to those two issues, and the change made in this PR was not the agreed-upon solution to those issues. (I don't know if we had even come to a consensus: I had argued for a different solution altogether.)

At this point, I suggest we just revert this part of https://github.com/riscv-non-isa/rvv-intrinsic-doc/pull/287.

topperc:
Neither gcc or LLVM implemented the original change from what I can tell.

Fixes #342

### DIFF
--- a/auto-generated/api-testing/vcpop.c
+++ b/auto-generated/api-testing/vcpop.c
@@ -1,58 +1,58 @@
 #include <riscv_vector.h>
 #include <stdint.h>
 
-unsigned int test_vcpop_m_b1(vbool1_t vs2, size_t vl) {
+unsigned long test_vcpop_m_b1(vbool1_t vs2, size_t vl) {
   return __riscv_vcpop_m_b1(vs2, vl);
 }
 
-unsigned int test_vcpop_m_b2(vbool2_t vs2, size_t vl) {
+unsigned long test_vcpop_m_b2(vbool2_t vs2, size_t vl) {
   return __riscv_vcpop_m_b2(vs2, vl);
 }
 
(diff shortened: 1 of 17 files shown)


nick-knight (approved):
LGTM, thanks @topperc.

topperc:
@kito-cheng @rofirrim can I merge this?

rofirrim:
> @kito-cheng @rofirrim can I merge this?

No objection from me.

rofirrim (approved):

rofirrim:
We want to cherrypick this into v1.0.x, don't we?

I can do that.

PR #344, opened 2024-06-11
Update CI configs

jerryzj:
- The build PDF is not appropriate in the PR stage as it is not aware of the latest document generator change
- This PR uses existed build generator.yaml to do the following checks
  - pylint
  - type-check
  - yapf-check
- The build python version is switch to py3.9/3.10/3.11

Signed-off-by: Jerry Zhang Jian <jerry.zhangjian@sifive.com>

### DIFF
--- a/.github/workflows/build-pdf.yml
+++ b/.github/workflows/build-pdf.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - main
-  pull_request:
   release:
     types:
       - created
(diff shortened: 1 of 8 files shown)


kito-cheng (approved):

jerryzj:
The pylint,type-check, and yapf-check are PASS now but the diff-autogen is failed.
I think we should merge this PR first and fix the document outdated issue later.

PR #346, opened 2024-06-13
makefile: let replace_float compatible with macOS builtin BSD sed

jerryzj:


kito-cheng (approved):
I don't have macOS, so I assume this should work correctly on that :P

PR #352, opened 2024-08-02
Address latest comment from ARC review

kito-cheng:
Review comment:

```
- "The value of the vxsat after a fixed-point intrinsic is UNSPECIFIED": Agreed, but this statement needs to be normative, and we need to be clear about exactly which intrinsics this applies to.  It seems to me that the intrinsics in Section A.4.1, A.4.3, and A.4.5 is the relevant set.  So, we ask that you add the following normative sentence to the top of each of those three sections: "After executing an intrinsic in this section, the vxsat CSR assumes an UNSPECIFIED value."

- "This includes the order in which the flag vxsat is updated in a program that executes a sequence of fixed-point intrinsics": This sentence should be deleted.

- Please add a non-normative note about future plans for vxsat.  As a starting point, you could replace the existing non-normative text with the following non-normative text: "This specification does not provide support for manipulating the vxsat CSR.  Since vxsat is not needed by a large majority of fixed-point code, we believe this specification is broadly useful as-is.  Nevertheless, we expect that a future extension will define an additional set of fixed-point intrinsics that update vxsat in a specified manner, along with intrinsics to explicitly read and write vxsat.  These new intrinsics would be interoperable with the intrinsics in this specification."
```

---

NOTE: This will backport to v1.0.x once merged into main branch.

### DIFF
--- a/auto-generated/intrinsic_funcs.adoc
+++ b/auto-generated/intrinsic_funcs.adoc
@@ -35506,6 +35506,7 @@ vuint64m8_t __riscv_vmv_v_x_u64m8(uint64_t rs1, size_t vl);
 
 [[vector-single-width-saturating-add-and-subtract]]
 ==== Vector Single-Width Saturating Add and Subtract Intrinsics
+After executing an intrinsic in this section, the `vxsat` CSR assumes an UNSPECIFIED value.
 
 [,c]
 ----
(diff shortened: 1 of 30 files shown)


kito-cheng:
PDF file here, this could save you 1~2 hours to generating that :P
[v-intrinsic-spec.pdf](https://github.com/user-attachments/files/16468286/v-intrinsic-spec.pdf)

rofirrim (approved):
LGTM. Thanks @kito-cheng

4vtomat (approved):
LGTM~

PR #355, opened 2024-08-02
Add custom llvm header support in generator

4vtomat:
It might be more extensions in the future, it's better to make each extension specify its own header file to prevent complex rule to check which header to use especially in llvm api tests.

### DIFF
--- a/rvv-intrinsic-generator/rvv_intrinsic_gen/generator.py
+++ b/rvv-intrinsic-generator/rvv_intrinsic_gen/generator.py
@@ -460,7 +460,7 @@ class APITestGenerator(Generator):
   def inst_group_epilogue(self):
     return ""
 
-  def write_file_header(self, has_float_type, name):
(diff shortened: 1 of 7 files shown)


rvv-intrinsic-generator/rvv_intrinsic_gen/generator.py
```diff
@@ -456,7 +460,7 @@ def inst_group_prologue(self):
   def inst_group_epilogue(self):
     return ""
 
-  def write_file_header(self, has_float_type, has_bfloat16_type, name):
+  def write_file_header(self, has_float_type, name):
```
jerryzj, line 463:
```suggestion
  def write_file_header(self, has_float_type):
```

jerryzj: The argument `name` is not used

PR #362, opened 2024-08-28
generator: add required extension info

jerryzj:
- Add a new field to store required extensions into a list
- Add required extension information for Zvk extensions
- Generate LLVM API test header dynamically based on required extensions
- Replace hardcoded LLVM header for Zvk and dynamically generate it
- Replaces #355 

Signed-off-by: Jerry Zhang Jian <jerry.zhangjian@sifive.com>

### DIFF
--- a/auto-generated/vector-crypto/llvm-api-tests/vctz.c
+++ b/auto-generated/vector-crypto/llvm-api-tests/vctz.c
@@ -1,12 +1,8 @@
 // REQUIRES: riscv-registered-target
-// RUN: %clang_cc1 -triple riscv64 -target-feature +v -target-feature +zvl512b \
+// RUN: %clang_cc1 -triple riscv64 -disable-O0-optnone \
 // RUN:   -target-feature +zvbb \
-// RUN:   -target-feature +zvbc \
-// RUN:   -target-feature +zvkg \
-// RUN:   -target-feature +zvkned \
-// RUN:   -target-feature +zvknhb \
-// RUN:   -target-feature +zvksed \
-// RUN:   -target-feature +zvksh -disable-O0-optnone \
+// RUN:   -target-feature +zve64x \
+// RUN:   -target-feature +experimental \
 // RUN:   -emit-llvm %s -o - | opt -S -passes=mem2reg | \
 // RUN:   FileCheck --check-prefix=CHECK-RV64 %s
 
--- a/rvv-intrinsic-generator/rvv_intrinsic_gen/enums.py
+++ b/rvv-intrinsic-generator/rvv_intrinsic_gen/enums.py
@@ -145,7 +145,8 @@ class InstInfo:
                inst_type=InstType.UNKNOWN,
                mem_type=MemType.NO_MEM,
                extra_attr=ExtraAttr.NO_ATTR,
-               NF=1):
+               NF=1,
+               required_ext=None):
     #pylint: disable=invalid-name
     self.SEW = SEW
     self.LMUL = LMUL
@@ -154,6 +155,9 @@ class InstInfo:
     self.mem_type = mem_type
     self.extra_attr = extra_attr
     self.NF = NF
+    if required_ext is None:
+      required_ext = []
+    self.required_ext = sorted(required_ext)
 
   def load_p(self):
     return self.mem_type == MemType.LOAD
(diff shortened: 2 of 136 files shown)


rvv-intrinsic-generator/rvv_intrinsic_gen/enums.py
```diff
@@ -154,6 +155,7 @@ def __init__(self,
     self.mem_type = mem_type
     self.extra_attr = extra_attr
     self.NF = NF
+    self.required_ext = required_ext
```
kito-cheng, line 158:
```suggestion
    if required_ext is None:
        required_ext = []
    self.required_ext = required_ext
```

rvv-intrinsic-generator/rvv_intrinsic_gen/enums.py
```diff
@@ -185,3 +187,16 @@ def get(args,
...
+
+  def add_required_ext(self, ext: str) -> None:
+    if self.required_ext is None:
+      self.required_ext = []
+    if ext not in self.required_ext:
+      self.required_ext.append(ext)
+
+  def remove_required_ext(self, ext: str) -> None:
+    if self.required_ext is not None and ext in self.required_ext:
+      self.required_ext.remove(ext)
```
kito-cheng, line 202: Then this part can be simplified

rvv-intrinsic-generator/rvv_intrinsic_gen/vector_crypto_inst.py
```diff
@@ -20,7 +20,8 @@ def gen(g):
       UITYPE,
       SEWS,
       LMULS,
-      decorators.has_masking_maskedoff_policy)
+      decorators.has_masking_maskedoff_policy,
+      required_ext_list=["v", "zvkb"])
```
kito-cheng, line 24: https://github.com/riscv/riscv-crypto/blob/main/doc/vector/riscv-crypto-spec-vector.adoc#1-extensions-overview

it should only require zve32x only? and same for most other vector crypto instruction

nibrunieAtSi5: Zvknhb would require SEW=64 (so ELEN >= 64) so I think it would imply more restrictions

jerryzj: I am going to add the `zve*` requirement in the runtime, which will make it easier to handle different SEW variation cases.

```
if int(args["SEW"] / lmul_num) == 64:
    inst_info.add_required_ext("zve64x")
else:
    inst_info.add_required_ext("zve32x")
```

kito-cheng (approved):

auto-generated/vector-crypto/llvm-api-tests/vctz.c
```diff
@@ -1,12 +1,9 @@
...
+// RUN: %clang_cc1 -triple riscv64 -disable-O0-optnone \
 // RUN:   -target-feature +zvbb \
-// RUN:   -target-feature +zvbc \
-// RUN:   -target-feature +zvkg \
-// RUN:   -target-feature +zvkned \
-// RUN:   -target-feature +zvknhb \
-// RUN:   -target-feature +zvksed \
-// RUN:   -target-feature +zvksh -disable-O0-optnone \
+// RUN:   -target-feature +zve64x \
+// RUN:   -target-feature +zvl256b \
```
topperc, line 5: Why does this require zvl256b?

rvv-intrinsic-generator/rvv_intrinsic_gen/enums.py
```diff
@@ -169,19 +173,66 @@ def get(args,
...
           inst_type,
           mem_type=MemType.NO_MEM,
-          extra_attr=ExtraAttr.NO_ATTR):
+          extra_attr=ExtraAttr.NO_ATTR,
+          required_ext=None):
     if decorator is None:
       # vsetvl and vsetvlmax
-      return InstInfo(args["SEW"], args["LMUL"], args["OP"], inst_type,
-                      mem_type, extra_attr)
+      if required_ext:
```
topperc, line 180: My python is pretty rusty. If required_exts is None, can we just pass it on to `InstInfo` without checking whether it is None or not?

jerryzj: Yes, I think so.

auto-generated/vector-crypto/llvm-api-tests/vctz.c
jerryzj: I think you are right, we don't need it.
The `zvl` constraint is generated by `EGW / LMUL` , but it is not applicable on `Zvbb` intrinsics, I will try if I can find a way to handle this.

jerryzj: Fixed, thanks.

rvv-intrinsic-generator/rvv_intrinsic_gen/enums.py
jerryzj: Updated and should be fine, thanks.

topperc (approved):
LGTM

PR #366, opened 2024-09-04
Import CONTRIBUTING.md from docs-spec-template

kito-cheng:
Sync with the docs-spec-template, and it request from the ratificaiton flow, which checking the document structure.


kito-cheng:
Gonna merge since this should be harmless to doc itself

PR #384, opened 2024-12-10
add dependabot to create PR to update the submodules to latest

kbroch-rivosinc:
relates to https://github.com/riscv-admin/docs-sig/issues/63


kito-cheng (approved):
Thanks :)

PR #394, opened 2025-04-07
Support Zvqdotq

4vtomat:
spec: https://github.com/riscv/riscv-dot-product


auto-generated/api-testing/vqdot.c
```diff
@@ -0,0 +1,102 @@
+#include <riscv_vector.h>
+#include <stdint.h>
+
+vint32mf2_t test_vqdot_vv_i32mf2(vint32mf2_t vd, vint8mf2_t vs2, vint8mf2_t vs1,
+                                 size_t vl) {
+  return __riscv_vqdot_vv_i32mf2(vd, vs2, vs1, vl);
+}
+
+vint32mf2_t test_vqdot_vx_i32mf2(vint32mf2_t vd, vint8mf2_t vs2, int32_t rs1,
```
topperc, line 9: I'm not sure if we should use `int32_t` for the scalar type here. There's really 4 sign bits, one for each 8-bit value. `uint32_t` may be more appropriate here.

topperc: @kito-cheng what's your opinion?

PR #399, opened 2025-06-16
Proposal for handling ambigious intrinsics for bfloat16 in zvfbfa

4vtomat:
TL;DR
    Propose a new rule for handling ambiguous bfloat16 intrinsics in zvfbfa

    As bfloat16 extension introduced in
    zvfbfa where there are some instructions such as widening arithmetic instructions, widening conversion instructions as well as narrowing conversion instructions will have intrinsic conflicts when we follow the rules in spec. Thus we'll need to expand the rules to accomodate those new coming intrinsics in the future.

    For non-overloaded intrinsics, adding additional type suffix before output type suffix to differentiate between ambiguous intrinsics, e.g.
    f16:  vfloat32m2 __riscv_vfwadd_f32m2(vfloat16m1_t a, vfloat16m1_t b, unsigned long vl);
    bf16: vfloat32m2 __riscv_vfwadd_bf16m1_f32m2(vbfloat16m1_t a, vbfloat16m1_t b, unsigned long vl);

    For overloaded intrinsics, adding additional suffix "_bf16" after name to differentiate between ambiguous intrinsics, e.g.
    f16:  vfloat16mf4_t __riscv_vfwcvt_f(vint8mf8_t vs2, size_t vl);
    bf16: vbfloat16mf4_t __riscv_vfwcvt_f_bf16(vint8mf8_t vs2, size_t vl);

    List of ambiguous intrinsics in Zvfbfa:

    Ambiguous on non-overloaded:

    Vector Widening BFloat Add/Subtract Intrinsics
    vfloat32mf2_t __riscv_vfwadd_vv_f32mf2(vbfloat16mf4_t vs2, vbfloat16mf4_t vs1, size_t vl);
    -> vfloat32mf2_t __riscv_vfwadd_vv_bf16mf4_f32mf2(vbfloat16mf4_t vs2, vbfloat16mf4_t vs1, size_t vl);
    vfloat32mf2_t __riscv_vfwsub_vv_f32mf2(vbfloat16mf4_t vs2, vbfloat16mf4_t vs1, size_t vl);
    -> vfloat32mf2_t __riscv_vfwsub_vv_bf16mf4_f32mf2(vbfloat16mf4_t vs2, vbfloat16mf4_t vs1, size_t vl);

    Vector Widening BFloat Multiply Intrinsics
    vfloat32mf2_t __riscv_vfwmul_vv_f32mf2(vbfloat16mf4_t vs2, vbfloat16mf4_t vs1, size_t vl);
    -> vfloat32mf2_t __riscv_vfwmul_vv_bf16mf4_f32mf2(vbfloat16mf4_t vs2, vbfloat16mf4_t vs1, size_t vl);

    Vector Widening BFloat Fused Multiply-Add Intrinsics
    vfloat32mf2_t __riscv_vfwmacc_vv_f32mf2(vfloat32mf2_t vd, vbfloat16mf4_t vs1, vbfloat16mf4_t vs2, size_t vl);
    -> vfloat32mf2_t __riscv_vfwmacc_vv_bf16mf4_f32mf2(vfloat32mf2_t vd, vbfloat16mf4_t vs1, vbfloat16mf4_t vs2, size_t vl);
    vfloat32mf2_t __riscv_vfwnmacc_vv_f32mf2(vfloat32mf2_t vd, vbfloat16mf4_t vs1, vbfloat16mf4_t vs2, size_t vl);
    -> vfloat32mf2_t __riscv_vfwnmacc_vv_bf16mf4_f32mf2(vfloat32mf2_t vd, vbfloat16mf4_t vs1, vbfloat16mf4_t vs2, size_t vl);
    vfloat32mf2_t __riscv_vfwmsac_vv_f32mf2(vfloat32mf2_t vd, vbfloat16mf4_t vs1, vbfloat16mf4_t vs2, size_t vl);
    -> vfloat32mf2_t __riscv_vfwmsac_vv_bf16mf4_f32mf2(vfloat32mf2_t vd, vbfloat16mf4_t vs1, vbfloat16mf4_t vs2, size_t vl);
    vfloat32mf2_t __riscv_vfwnmsac_vv_f32mf2(vfloat32mf2_t vd, vbfloat16mf4_t vs1, vbfloat16mf4_t vs2, size_t vl);
    -> vfloat32mf2_t __riscv_vfwnmsac_vv_bf16mf4_f32mf2(vfloat32mf2_t vd, vbfloat16mf4_t vs1, vbfloat16mf4_t vs2, size_t vl);

    Widening BFloat Type-Convert Intrinsics
    vfloat32mf2_t __riscv_vfwcvt_f_f_v_f32mf2(vbfloat16mf4_t vs2, size_t vl);
    -> vfloat32mf2_t __riscv_vfwcvt_f_f_v_bf16mf4_f32mf2(vbfloat16mf4_t vs2, size_t vl);

    Narrowing BFloat Type-Convert Intrinsics
    vint8mf8_t __riscv_vfncvt_x_f_w_i8mf8(vbfloat16mf4_t vs2, size_t vl);
    -> vint8mf8_t __riscv_vfncvt_x_f_w_bf16mf4_i8mf8(vbfloat16mf4_t vs2, size_t vl);
    vint8mf8_t __riscv_vfncvt_rtz_x_f_w_i8mf8(vbfloat16mf4_t vs2, size_t vl);
    -> vint8mf8_t __riscv_vfncvt_rtz_x_f_w_bf16mf4_i8mf8(vbfloat16mf4_t vs2, size_t vl);

    Vector BFloat Classify Intrinsics
    vuint16mf4_t __riscv_vfclass_v_u16mf4(vbfloat16mf4_t vs2, size_t vl);
    -> vuint16mf4_t __riscv_vfclass_v_bf16mf4_u16mf4(vbfloat16mf4_t vs2, size_t vl);

    Ambiguous on overloaded:

    Widening BFloat Type-Convert Intrinsics
    vbfloat16mf4_t __riscv_vfwcvt_f(vint8mf8_t vs2, size_t vl);
    -> vbfloat16mf4_t __riscv_vfwcvt_f_bf16(vint8mf8_t vs2, size_t vl);

    Narrowing BFloat Type-Convert Intrinsics
    vbfloat16mf4_t __riscv_vfncvt_f(vfloat32mf2_t vs2, size_t vl);
    -> vbfloat16mf4_t __riscv_vfncvt_f_bf16(vfloat32mf2_t vs2, size_t vl);
    vbfloat16mf4_t __riscv_vfncvt_rod_f(vfloat32mf2_t vs2, size_t vl);
    -> vbfloat16mf4_t __riscv_vfncvt_rod_f_bf16(vfloat32mf2_t vs2, size_t vl);

### DIFF
--- a/doc/rvv-intrinsic-spec.adoc
+++ b/doc/rvv-intrinsic-spec.adoc
@@ -297,6 +297,56 @@ vbool8_t __riscv_vreinterpret_v_i32m1_b8 (vint32m1_t src);
 vbool4_t __riscv_vreinterpret_v_i32m1_b4 (vint32m1_t src);
 ----
 
+=== Vector BFloat instructions
+
+For vector bfloat instructions, appending additional input type before output type to disambiguate between `zvfhmin` instructions, e.g. `vfloat32mf2_t __riscv_vfwadd_vv_f32mf2(vfloat16mf4_t vs2, vfloat16mf4_t vs1, size_t vl);`.
+
+==== Vector Widening BFloat Add instruction
+[,c]
+----
+vfloat32mf2_t __riscv_vfwadd_vv_bf16mf4_f32mf2(vbfloat16mf4_t vs2, vbfloat16mf4_t vs1, size_t vl);
+----
+
+==== Vector Widening BFloat Sub instruction
+[,c]
+----
+vfloat32mf2_t __riscv_vfwsub_vv_bf16mf4_f32mf2(vbfloat16mf4_t vs2, vbfloat16mf4_t vs1, size_t vl);
+----
+
+==== Vector Widening BFloat Mul instruction
+[,c]
+----
+vfloat32mf2_t __riscv_vfwmul_vv_bf16mf4_f32mf2(vbfloat16mf4_t vs2, vbfloat16mf4_t vs1, size_t vl);
+----
+
+==== Vector Widening BFloat Fused Multiply-Add instructions
+[,c]
+----
+vfloat32mf2_t __riscv_vfwmacc_vv_bf16mf4_f32mf2(vfloat32mf2_t vd, vbfloat16mf4_t vs1, vbfloat16mf4_t vs2, size_t vl);
+vfloat32mf2_t __riscv_vfwnmacc_vv_bf16mf4_f32mf2(vfloat32mf2_t vd, vbfloat16mf4_t vs1, vbfloat16mf4_t vs2, size_t vl);
+vfloat32mf2_t __riscv_vfwmsac_vv_bf16mf4_f32mf2(vfloat32mf2_t vd, vbfloat16mf4_t vs1, vbfloat16mf4_t vs2, size_t vl);
+vfloat32mf2_t __riscv_vfwnmsac_vv_bf16mf4_f32mf2(vfloat32mf2_t vd, vbfloat16mf4_t vs1, vbfloat16mf4_t vs2, size_t vl);
+----
+
+==== Vector Widening BFloat Type-Convert instruction
+[,c]
+----
+vfloat32mf2_t __riscv_vfwcvt_f_f_v_bf16mf4_f32mf2(vbfloat16mf4_t vs2, size_t vl);
+----
+
+==== Vector Narrowing BFloat Type-Convert instruction
+[,c]
+----
+vint8mf8_t __riscv_vfncvt_x_f_w_bf16mf4_i8mf8(vbfloat16mf4_t vs2, size_t vl);
+vint8mf8_t __riscv_vfncvt_rtz_x_f_w_bf16mf4_i8mf8(vbfloat16mf4_t vs2, size_t vl);
+----
+
+==== Vector BFloat Classify instruction
+[,c]
+----
+vuint16mf4_t __riscv_vfclass_v_bf16mf4_u16mf4(vbfloat16mf4_t vs2, size_t vl);
+----
+
 [[implicit-naming-scheme]]
 === Implicit (Overloaded) naming scheme
 
@@ -362,6 +412,25 @@ vint32m1_t __riscv_vwadd_vv(vint16mf2_t vs2, vint16mf2_t vs1, size_t vl);
 vint32m1_t __riscv_vwadd_vx(vint16mf2_t vs2, int16_t rs1, size_t vl);
 vint32m1_t __riscv_vwadd_wv(vint32m1_t vs2, vint16mf2_t vs1, size_t vl);
 vint32m1_t __riscv_vwadd_wx(vint32m1_t vs2, int16_t rs1, size_t vl);
+----
+
+==== Widening BFloat Type-Convert instructions
+
+Add `_bf16` suffix to disambiguate between `zvfhmin` instructions, e.g. `vfloat16mf4_t __riscv_vfwcvt_f(vint8mf8_t vs2, size_t vl);`.
+
+[,c]
+----
+vbfloat16mf4_t __riscv_vfwcvt_f_bf16(vint8mf8_t vs2, size_t vl);
+----
+
+==== Narrowing BFloat Type-Convert instructions
+
+Add `_bf16` suffix to disambiguate between `zvfhmin` instructions, e.g. `vfloat16mf4_t __riscv_vfncvt_f(vfloat32mf2_t vs2, size_t vl);`.
+
+[,c]
+----
+vbfloat16mf4_t __riscv_vfncvt_f_bf16(vfloat32mf2_t vs2, size_t vl);
+vbfloat16mf4_t __riscv_vfncvt_rod_f_bf16(vfloat32mf2_t vs2, size_t vl);
 ----
 
 ==== Type-convert instructions


auto-generated/bfloat16/api-testing/vle16.c
```diff
@@ -1,53 +1,56 @@
 #include <riscv_vector.h>
 #include <stdint.h>
 
-vbfloat16mf4_t test_vle16_v_bf16mf4(const __bf16 *rs1, size_t vl) {
-  return __riscv_vle16_v_bf16mf4(rs1, vl);
+vbfloat16mf4_t test_vle16_v_bf16mf4_alt(const __bf16 *rs1, size_t vl) {
```
tclin914, line 4: Why does vle16 need the `_alt` suffix? Does it have a different return type from the fp16 version? `test_vle16_v_bf16mf4` vs. `test_vle16_v_f16mf4`

4vtomat: That's a good question! The reason why I also put `_alt` suffix to intrinsics that is not ambiguous is because of consistency. If we add the suffix to all of intrinsics that need bf16, then users might be less confused about the rule, otherwise we need to put them in exception list and users need to lookup when using them.
However I'm fine with both approach lol

kito-cheng: But this not following the rule "Propose a new rule for intrinsics that need vtype.altfmt to be set to 1."? vle only care about LMUL and SEW I think?

4vtomat:
> Is the proposal to add _alt to the ambiguous intrinsics that don't naturally have a bf16 type in their now or adding _alt to all bf16 intrinsics?
> 
> vle isn't ambiguous so I guess all intrinsics? But why wasn't vse or the other load intrinsics changed?

Yeah correct, this proposal adds `_alt` to all bf16 intrinsics that need AltFmt. I just add 1 test case because of ease of reviewing, but seems some bf16 intrinsics don't need to set AltFmt so I'm gonna update and add all of intrinsics that got affected by this proposal~

auto-generated/bfloat16/api-testing/vle16.c
4vtomat: Yeah, that's incorrect example, I'll update and put all of intrinsics that are changed by this proposal!

4vtomat:
> Description doesn't show or mention that the _alt goes before the policy suffixes.

Updated~

ebavier:
I have two concerns about this proposal of disambiguation by adding `_alt`:

1. The `vtype.altfmt` is an instruction encoding implementation detail that doesn't seem appropriate to let leak into a software API.
2. Confusion when `altfmt` is used for other types.  Consider the Zvfwldot16bf and Zvfqldot8f extension examples cited where `altfmt` is used to select the FP8 format.  While it's likely that someone writing intrinsics code might be referencing documentation and so will know what form of `_alt` function to use, it's also likely that someone reading the code might have difficulty determining what datatype `_alt` refers to in context.

I would prefer a different method for disambiguating bf16 (and fp8) intrinsics arguments, something more like the 1st approach mentioned but with the input type tag preceding the output type tag.  This would better align with the ordering used in other intrinsics (e.g. in `__riscv_vfredmax_vs_f32m8_f32m1` we have `__riscv_vfredmax_vs_{in}_{out}`).

Concretely:

- Vector Widening BFloat Add/Subtract Intrinsics
  - vfloat32mf2_t __riscv_vfwadd_vv_f32mf2(vbfloat16mf4_t vs2, vbfloat16mf4_t vs1, size_t vl);
-> vfloat32mf2_t __riscv_vfwadd_vv_bf16mf4_f32mf2(vbfloat16mf4_t vs2, vbfloat16mf4_t vs1, size_t vl);
  - vfloat32mf2_t __riscv_vfwsub_vv_f32mf2(vbfloat16mf4_t vs2, vbfloat16mf4_t vs1, size_t vl);
-> vfloat32mf2_t __riscv_vfwsub_vv_bf16mf4_f32mf2(vbfloat16mf4_t vs2, vbfloat16mf4_t vs1, size_t vl);
- Vector Widening BFloat Multiply Intrinsics
  - vfloat32mf2_t __riscv_vfwmul_vv_f32mf2(vbfloat16mf4_t vs2, vbfloat16mf4_t vs1, size_t vl);
-> vfloat32mf2_t __riscv_vfwmul_vv_bf16mf4_f32mf2(vbfloat16mf4_t vs2, vbfloat16mf4_t vs1, size_t vl);
- Vector Widening BFloat Fused Multiply-Add Intrinsics
  - vfloat32mf2_t __riscv_vfwmacc_vv_f32mf2(vfloat32mf2_t vd, vbfloat16mf4_t vs1, vbfloat16mf4_t vs2, size_t vl);
-> vfloat32mf2_t __riscv_vfwmacc_vv_bf16mf4_f32mf2(vfloat32mf2_t vd, vbfloat16mf4_t vs1, vbfloat16mf4_t vs2, size_t vl);
  - vfloat32mf2_t __riscv_vfwnmacc_vv_f32mf2(vfloat32mf2_t vd, vbfloat16mf4_t vs1, vbfloat16mf4_t vs2, size_t vl);
-> vfloat32mf2_t __riscv_vfwnmacc_vv_bf16mf4_f32mf2(vfloat32mf2_t vd, vbfloat16mf4_t vs1, vbfloat16mf4_t vs2, size_t vl);
  - vfloat32mf2_t __riscv_vfwmsac_vv_f32mf2(vfloat32mf2_t vd, vbfloat16mf4_t vs1, vbfloat16mf4_t vs2, size_t vl);
-> vfloat32mf2_t __riscv_vfwmsac_vv_bf16mf4_f32mf2(vfloat32mf2_t vd, vbfloat16mf4_t vs1, vbfloat16mf4_t vs2, size_t vl);
  - vfloat32mf2_t __riscv_vfwnmsac_vv_f32mf2(vfloat32mf2_t vd, vbfloat16mf4_t vs1, vbfloat16mf4_t vs2, size_t vl);
-> vfloat32mf2_t __riscv_vfwnmsac_vv_bf16mf4_f32mf2(vfloat32mf2_t vd, vbfloat16mf4_t vs1, vbfloat16mf4_t vs2, size_t vl);
- Widening BFloat Type-Convert Intrinsics
  - vfloat32mf2_t __riscv_vfwcvt_f_f_v_f32mf2(vbfloat16mf4_t vs2, size_t vl);
-> vfloat32mf2_t __riscv_vfwcvt_f_f_v_bf16mf4_f32mf2(vbfloat16mf4_t vs2, size_t vl);
- Narrowing BFloat Type-Convert Intrinsics
  - vint8mf8_t __riscv_vfncvt_x_f_w_i8mf8(vbfloat16mf4_t vs2, size_t vl);
-> vint8mf8_t __riscv_vfncvt_x_f_w_bf16mf4_i8mf8(vbfloat16mf4_t vs2, size_t vl);
  - vint8mf8_t __riscv_vfncvt_rtz_x_f_w_i8mf8(vbfloat16mf4_t vs2, size_t vl);
-> vint8mf8_t __riscv_vfncvt_rtz_x_f_w_bf16mf4_i8mf8(vbfloat16mf4_t vs2, size_t vl);
- Vector BFloat Classify Intrinsics
  - vuint16mf4_t __riscv_vfclass_v_u16mf4(vbfloat16mf4_t vs2, size_t vl);
-> vuint16mf4_t __riscv_vfclass_v_bf16mf4_u16mf4(vbfloat16mf4_t vs2, size_t vl);

Ambiguous on overloaded:

- Widening BFloat Type-Convert Intrinsics
  - vbfloat16mf4_t __riscv_vfwcvt_f(vint8mf8_t vs2, size_t vl);
-> vbfloat16mf4_t __riscv_vfwcvt_f_bf16mf4(vint8mf8_t vs2, size_t vl);
- Narrowing BFloat Type-Convert Intrinsics
  - vbfloat16mf4_t __riscv_vfncvt_f(vfloat32mf2_t vs2, size_t vl);
-> vbfloat16mf4_t __riscv_vfncvt_f_bf16mf4(vfloat32mf2_t vs2, size_t vl);
  - vbfloat16mf4_t __riscv_vfncvt_rod_f(vfloat32mf2_t vs2, size_t vl);
-> vbfloat16mf4_t __riscv_vfncvt_rod_f_bf16mf4(vfloat32mf2_t vs2, size_t vl);

As for intrinsics for the Zvfwldot16bf and Zvfqldot8f extensions, these are naturally treated as reductions, and, following existing reduction intrinsics, would have a type tag specifying the input argument type so would need no further disambiguation:

```
vint32m1_t __riscv_sf_vqldotu_vv_u8m4_i32m1(vuint8m4_t vs2, vint8m4_t vs1, size_t vl);
vint32m1_t __riscv_sf_vqldots_vv_i8m4_i32m1(vint8m4_t vs2, vint8m4_t vs1, size_t vl);

vfloat32m1_t __riscv_sf_vfwldot_vv_bf16m1_f32m1(vbfloat16m1_t vs2, vbfloat16m1_t vs1, size_t vl);
vfloat32m1_t __riscv_sf_vfwldot_vv_bf16m2_f32m1(vbfloat16m2_t vs2, vbfloat16m2_t vs1, size_t vl);
vfloat32m1_t __riscv_sf_vfwldot_vv_bf16m4_f32m1(vbfloat16m4_t vs2, vbfloat16m4_t vs1, size_t vl);
// etc..

// Maps to vfqldot.vv, altfmt=0:
vfloat32m1_t __riscv_sf_vfqldot_vv_e4m3m4_e4m3m4_f32m1(vfloate4m3m4_t vs1, vfloate4m3m4_t vs2, size_t vl);
// Maps to vfqldot.vv, altfmt=1:
vfloat32m1_t __riscv_sf_vfqldot_vv_e4m3m4_e5m2m4_f32m1(vfloate4m3m4_t vs1, vfloate5m2m4_t vs2, size_t vl);
// Maps to vfqldot.alt.vv, altfmt=0:
vfloat32m1_t __riscv_sf_vfqldot_vv_e5m2m4_e4m3m4_f32m1(vfloate5m2m4_t vs1, vfloate4m3m4_t vs2, size_t vl);
// Maps to vfqldot.alt.vv, altfmt=1:
vfloat32m1_t __riscv_sf_vfqldot_vv_e5m2m4_e5m2m4_f32m1(vfloate5m2m4_t vs1, vfloate5m2m4_t vs2, size_t vl);
```

Note I've made up the `vfloate4m3m4_t` etc types above (the standard mantissa-length encoding has an unfortunate similarity to the adjacent LMUL encoding); I don't know whether we plan to define proper types for the OFP8 and OFP4 formats or continue to use `vuint8m4_t` etc as stand-ins.

Anyhow, I think something like this is preferable; more informative in the presence of different `altfmt` contexts, and likely less error-prone, especially if OFP8 formats are type-punned from `uint8` types.

4vtomat:
Thanks Eric for reviewing this!
> I have two concerns about this proposal of disambiguation by adding `_alt`:
> 
> 1. The `vtype.altfmt` is an instruction encoding implementation detail that doesn't seem appropriate to let leak into a software API.

I think for RVV overloaded and policy intrinsic, it also appears that `vtype.vta` and `vtype.vma` are exposed as a part of API right? e.g. `vadd_tu`, `vadd_tumu`, etc.

> 2. Confusion when `altfmt` is used for other types.  Consider the Zvfwldot16bf and Zvfqldot8f extension examples cited where `altfmt` is used to select the FP8 format.  While it's likely that someone writing intrinsics code might be referencing documentation and so will know what form of `_alt` function to use, it's also likely that someone reading the code might have difficulty determining what datatype `_alt` refers to in context.

My bad, I should not confuse reviewers that this is the proposal specifically for differentiate between "bfloat16" and "non-bfloat16" intrinsics, it's supposed to be used to differentiate between intrinsics that "need" `vtype.altfmt` and intrinsics that "don't need" `vtype.altfmt` no matter what the type is.

> 
> I would prefer a different method for disambiguating bf16 (and fp8) intrinsics arguments, something more like the 1st approach mentioned but with the input type tag preceding the output type tag. This would better align with the ordering used in other intrinsics (e.g. in `__riscv_vfredmax_vs_f32m8_f32m1` we have `__riscv_vfredmax_vs_{in}_{out}`).
> 

That's also a good way to do so, but I'm a bit worrying about consistency on this approach, for example, bfloat16 version of `__riscv_vfwmacc_vv_f32mf2` will be `__riscv_vfwmacc_vv_bf16mf4_f32mf2` which is not consistent with other `vfwmacc_vv` intrinsic and should be listed as exception in [this](https://github.com/riscv-non-isa/rvv-intrinsic-doc/blob/main/doc/rvv-intrinsic-spec.adoc#exceptions-in-the-explicit-non-overloaded-naming-scheme) section, then all of intrinsic listed in the comments should be putting here which might not be expected. However if we make this a general rule for all intrinsic in [this](https://github.com/riscv-non-isa/rvv-intrinsic-doc/blob/main/doc/rvv-intrinsic-spec.adoc#explicit-non-overloaded-naming-scheme) section, then all of intrinsics can be formularized without any exception, that kind of minimizes the exception section.

Anyways I think it's worth it to have wide discussion about this and come up with the best solution!

4vtomat:
> I agree with @ebavier.
> 
> > I think for RVV overloaded and policy intrinsic, it also appears that vtype.vta and vtype.vma are exposed as a part of API right? e.g. vadd_tu, vadd_tumu, etc.
> 
> Those names are describing a behavior the architecture has. Not really the name of a particular bit. The altfmt bit in vtype is trying to do multiple things with a single bit. If it had been called the `bf16` bit would you be making a different argument?
> 
> Maybe a better comparison is that our intrinsic names use `f32m8`, `u32m8`, `i32m8` instead of `e32m8`. Floating point and integer are already determined by the opcode name so we could have just used SEW and LMUL, but we are more explicit. Though vadd_e32m8 would be ambiguous since we support signed/unsigned types. So this isn't a great comparison either.

Yeah that make sense to me, one thing I think we need to discuss is what's the better way to handle overloaded intrinsic, the type information might not be supposed to put in the naming, otherwise we'll have different name for each overloaded intrinsic and it's meaningless to have such overloaded intrinsics. So I think we can maybe have a general suffix for those ambiguous intrinsic, for example, "_bf", "_f8" something like that rather than using type.

4vtomat:
> > for example, "_bf", "_f8" something like that
> 
> Yes, right, the overloaded intrinsics should not have encoded lmul in their names, which can be inferred from the input types, and a general suffix is enough to disambiguate.
> 
> I think `_bf16` should be used in place of `_bf`. I like the shorter `_bf`, but `_bf16` is closer to what's used everywhere else.
> 
> Ambiguous on overloaded:
> 
> * Widening BFloat Type-Convert Intrinsics
>   
>   * vbfloat16mf4_t __riscv_vfwcvt_f(vint8mf8_t vs2, size_t vl);
>     -> vbfloat16mf4_t __riscv_vfwcvt_f_bf16(vint8mf8_t vs2, size_t vl);
>   * vbfloat16m8_t __riscv_vfwcvt_f(vint8m4_t vs2, size_t vl);
>     -> vbfloat16m8_t __riscv_vfwcvt_f_bf16(vint8m4_t vs2, size_t vl);
> * Narrowing BFloat Type-Convert Intrinsics
>   
>   * vbfloat16mf4_t __riscv_vfncvt_f(vfloat32mf2_t vs2, size_t vl);
>     -> vbfloat16mf4_t __riscv_vfncvt_f_bf16(vfloat32mf2_t vs2, size_t vl);
>   * vbfloat16mf4_t __riscv_vfncvt_rod_f(vfloat32mf2_t vs2, size_t vl);
>     -> vbfloat16mf4_t __riscv_vfncvt_rod_f_bf16(vfloat32mf2_t vs2, size_t vl);

Do you think it's also good for this to apply to non-overloaded intrinsics? e.g. 
```
vfloat32mf2_t __riscv_vfwadd_vv_f32mf2(vbfloat16mf4_t vs2, vbfloat16mf4_t vs1, size_t vl);
-> vfloat32mf2_t __riscv_vfwadd_vv_bf_f32mf2(vbfloat16mf4_t vs2, vbfloat16mf4_t vs1, size_t vl);
```
I think it's more consistent with overloaded intrinsic and we also don't need to add another type suffix.

ebavier:
> Do you think it's also good for this to apply to non-overloaded intrinsics?

I would be less worried about apply this to non-overloaded intrinsics because, e.g. the reduction intrinsics all have "m1" lmul tags in the return type, even though that's not strictly necessary.  I suppose in a way it makes it incredibly obvious that the returned value is a vector and not a scalar.

4vtomat:
> > Do you think it's also good for this to apply to non-overloaded intrinsics?
> 
> I would be less worried about apply this to non-overloaded intrinsics because, e.g. the reduction intrinsics all have "m1" lmul tags in the return type, even though that's not strictly necessary. I suppose in a way it makes it incredibly obvious that the returned value is a vector and not a scalar.

So do you mean we can keep non-overloaded one like
```
vfloat32mf2_t __riscv_vfwadd_vv_f32mf2(vbfloat16mf4_t vs2, vbfloat16mf4_t vs1, size_t vl);
-> vfloat32mf2_t __riscv_vfwadd_vv_bf16mf4_f32mf2(vbfloat16mf4_t vs2, vbfloat16mf4_t vs1, size_t vl);
```
and make overloaded one like
```
vfloat32mf2_t __riscv_vfwadd_vv_f32mf2(vbfloat16mf4_t vs2, vbfloat16mf4_t vs1, size_t vl);
-> vfloat32mf2_t __riscv_vfwadd_vv_bf(vbfloat16mf4_t vs2, vbfloat16mf4_t vs1, size_t vl);
```

ebavier:
> Should we drop mf4 from `__riscv_vfwadd_vv_bf16mf4_f32mf2` leaving `__riscv_vfwadd_vv_bf16_` so it matches the overload name?

In my opinion, no.  The lmul component makes it clear the type refers to a vector.

4vtomat:
Updated proposal to be `zvfbfa` specific and applied the rule we discussed

topperc (approved):
LGTM

PR #414, opened 2025-08-31
Add vfext to get_overloaded_op_name

4vtomat:
vfext.vf2 is introduced here: https://github.com/aswaterman/riscv-misc/blob/main/isa/zvfofp4min.adoc it's possible that there are vfext.vf4, vfext.vf8 added so we need a distinct name for them.


topperc (approved):
LGTM

PR #415, opened 2025-10-01
Updated rvv_matmul.c to clarify that B is transposed before multiplication

aznszn:
From reading the code it is not apparent that the golden and actual matmul functions are not doing standard matrix multiplication but rather expect B to be already transposed.

This PR adds comments and renames b to b_t to clarify this


eopXD (approved):
LGTM

PR #418, opened 2026-01-22
Zvqdotq zvdot4a8i renaming

nibrunieAtSi5:
Implementing renaming from **Zvqdotq** to **Zvodt4a8i** mandated after the ARC review on the fast track.

(**riscv-isa-manual** PR: https://github.com/riscv/riscv-isa-manual/pull/2576)

This PR is made out of two commits:
- bba02b9aadbd098a49b7f4ccde9b49411be890b5 modify instruction and template
- 9579b4a488d3d4368e19a395cfb5ff988f3312ba imports the auto-generated change

One of the thing I had to do is modify is to add `"dot4a"` in the list of prefix checked-for when determining if an instruction is quad widening (`vdot4a*` are quad widening in some way although the prefix `q` is not in the mnemonic).


4vtomat (approved):
LGTM~

PR #422, opened 2026-01-29
Use [u]int32 vector for input arguments for zvdot4a8i

4vtomat:
The instruction is defined for SEW=32, the most optimized way is to use
with vle32 to prevent vsetvli toggling.

### DIFF
--- a/auto-generated/api-testing/vdot4aus.c
+++ b/auto-generated/api-testing/vdot4aus.c
@@ -1,53 +1,53 @@
 #include <riscv_vector.h>
 #include <stdint.h>
 
-vint32mf2_t test_vdot4aus_vx_i32mf2(vint32mf2_t vd, vuint8mf2_t vs2,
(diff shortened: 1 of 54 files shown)


auto-generated/api-testing/vdot4aus.c
```diff
@@ -1,53 +1,53 @@
 #include <riscv_vector.h>
 #include <stdint.h>
 
-vint32mf2_t test_vdot4aus_vx_i32mf2(vint32mf2_t vd, vuint8mf2_t vs2,
+vint32mf2_t test_vdot4aus_vx_u32mf2(vuint32mf2_t vd, vuint32mf2_t vs2,
```
topperc: The type of vd and the return type should match. It's a tied opeand.

topperc (approved):
LGTM, but wait for @nibrunieAtSi5

PR #428, opened 2026-03-12
Support zvbdota

4vtomat:
spec: https://github.com/aswaterman/riscv-misc/blob/main/isa/ldot-bdot/ldot-bdot.adoc#zvbdota-family-of-batched-dot-product-extensions-version-02

stacked on: https://github.com/riscv-non-isa/riscv-rvv-intrinsic-doc/pull/427

### DIFF
--- a/auto-generated/api-testing/vfbdota.c
+++ b/auto-generated/api-testing/vfbdota.c
@@ -1,0 +1,94 @@
+#include <riscv_vector.h>
+#include <stdint.h>
+
+vfloat32m1_t test_vfbdota_vv_f32m1(vfloat32m1_t vd, vfloat32m8_t vs2,
+                                   vfloat32m1_t vs1, size_t vl) {
+  return __riscv_vfbdota_vv_f32m1(vd, vs2, vs1, 0, vl);
+}
+
+vfloat32m2_t test_vfbdota_vv_f32m2(vfloat32m2_t vd, vfloat32m8_t vs2,
+                                   vfloat32m1_t vs1, size_t vl) {
+  return __riscv_vfbdota_vv_f32m2(vd, vs2, vs1, 0, vl);
+}
+
+vfloat32m4_t test_vfbdota_vv_f32m4(vfloat32m4_t vd, vfloat32m8_t vs2,
+                                   vfloat32m1_t vs1, size_t vl) {
+  return __riscv_vfbdota_vv_f32m4(vd, vs2, vs1, 0, vl);
+}
+
+vfloat32m8_t test_vfbdota_vv_f32m8(vfloat32m8_t vd, vfloat32m8_t vs2,
+                                   vfloat32m1_t vs1, size_t vl) {
+  return __riscv_vfbdota_vv_f32m8(vd, vs2, vs1, 0, vl);
+}
+
+vfloat32m1_t test_vfbdota_vv_f32m1_m(vbool32_t vm, vfloat32m1_t vd,
+                                     vfloat32m8_t vs2, vfloat32m1_t vs1,
+                                     size_t vl) {
+  return __riscv_vfbdota_vv_f32m1_m(vm, vd, vs2, vs1, 0, vl);
+}
+
+vfloat32m2_t test_vfbdota_vv_f32m2_m(vbool16_t vm, vfloat32m2_t vd,
+                                     vfloat32m8_t vs2, vfloat32m1_t vs1,
+                                     size_t vl) {
+  return __riscv_vfbdota_vv_f32m2_m(vm, vd, vs2, vs1, 0, vl);
+}
+
+vfloat32m4_t test_vfbdota_vv_f32m4_m(vbool8_t vm, vfloat32m4_t vd,
+                                     vfloat32m8_t vs2, vfloat32m1_t vs1,
+                                     size_t vl) {
+  return __riscv_vfbdota_vv_f32m4_m(vm, vd, vs2, vs1, 0, vl);
+}
+
+vfloat32m8_t test_vfbdota_vv_f32m8_m(vbool4_t vm, vfloat32m8_t vd,
+                                     vfloat32m8_t vs2, vfloat32m1_t vs1,
+                                     size_t vl) {
+  return __riscv_vfbdota_vv_f32m8_m(vm, vd, vs2, vs1, 0, vl);
+}
+
+vfloat32m1_t test_vfbdota_vv_f32m1_rm(vfloat32m1_t vd, vfloat32m8_t vs2,
+                                      vfloat32m1_t vs1, size_t vl) {
+  return __riscv_vfbdota_vv_f32m1_rm(vd, vs2, vs1, 0, __RISCV_FRM_RNE, vl);
+}
+
+vfloat32m2_t test_vfbdota_vv_f32m2_rm(vfloat32m2_t vd, vfloat32m8_t vs2,
+                                      vfloat32m1_t vs1, size_t vl) {
+  return __riscv_vfbdota_vv_f32m2_rm(vd, vs2, vs1, 0, __RISCV_FRM_RNE, vl);
+}
+
+vfloat32m4_t test_vfbdota_vv_f32m4_rm(vfloat32m4_t vd, vfloat32m8_t vs2,
+                                      vfloat32m1_t vs1, size_t vl) {
+  return __riscv_vfbdota_vv_f32m4_rm(vd, vs2, vs1, 0, __RISCV_FRM_RNE, vl);
+}
+
+vfloat32m8_t test_vfbdota_vv_f32m8_rm(vfloat32m8_t vd, vfloat32m8_t vs2,
+                                      vfloat32m1_t vs1, size_t vl) {
+  return __riscv_vfbdota_vv_f32m8_rm(vd, vs2, vs1, 0, __RISCV_FRM_RNE, vl);
+}
+
+vfloat32m1_t test_vfbdota_vv_f32m1_rm_m(vbool32_t vm, vfloat32m1_t vd,
+                                        vfloat32m8_t vs2, vfloat32m1_t vs1,
+                                        size_t vl) {
+  return __riscv_vfbdota_vv_f32m1_rm_m(vm, vd, vs2, vs1, 0, __RISCV_FRM_RNE,
+                                       vl);
+}
+
+vfloat32m2_t test_vfbdota_vv_f32m2_rm_m(vbool16_t vm, vfloat32m2_t vd,
+                                        vfloat32m8_t vs2, vfloat32m1_t vs1,
+                                        size_t vl) {
+  return __riscv_vfbdota_vv_f32m2_rm_m(vm, vd, vs2, vs1, 0, __RISCV_FRM_RNE,
+                                       vl);
+}
+
+vfloat32m4_t test_vfbdota_vv_f32m4_rm_m(vbool8_t vm, vfloat32m4_t vd,
+                                        vfloat32m8_t vs2, vfloat32m1_t vs1,
+                                        size_t vl) {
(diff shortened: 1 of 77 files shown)


auto-generated/api-testing/vfbdota.c
```diff
@@ -0,0 +1,123 @@
+#include <riscv_vector.h>
+#include <stdint.h>
+
+vfloat32mf2_t test_vfbdota_vv_f32m1_f32mf2(vfloat32mf2_t vd, vfloat32m8_t vs2,
```
topperc: Do we need to support fractional destinations? EMUL is determined by `ceil(8*EEW/VLEN)`. So that should always be >= 1.

auto-generated/api-testing/vfqwbdota.c
```diff
@@ -0,0 +1,258 @@
...
+  return __riscv_vfqwbdota_vv_f8e5m2m1_f32m4_m(vm, vd, vs2, vs1, 0, vl);
+}
+
+vfloat32m8_t test_vfqwbdota_vv_f8e5m2m1_f32m8_m(vbool4_t vm, vfloat32m8_t vd,
+                                                vuint8m8_t vs2, vuint8m1_t vs1,
+                                                size_t vl) {
+  return __riscv_vfqwbdota_vv_f8e5m2m1_f32m8_m(vm, vd, vs2, vs1, 0, vl);
+}
+
+vfloat32mf2_t test_vfqwbdota_vv_f8e4m3m1_f32mf2_rm(vfloat32mf2_t vd,
```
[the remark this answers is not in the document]
4vtomat: Got it, thanks!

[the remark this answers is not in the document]
4vtomat: btw do we intend to support f32 for long dot product?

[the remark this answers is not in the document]
aswaterman: No, it has not been specified (on the basis that it is not expected to be much faster in practice than VFMUL + VFREDSUM).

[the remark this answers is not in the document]
4vtomat: I see, thanks for clarification!

topperc (approved):
LGTM

topperc (approved):

rvv-intrinsic-generator/rvv_intrinsic_gen/generator.py
```diff
@@ -295,9 +295,11 @@ def get_overloaded_op_name(name):
...
     elif any(op in name for op in ["vlmul_ext", "vlmul_trunc"]):
       overloaded_name = "_".join([sn[0], sn[1], sn[-1]])
-    elif any(op in name for op in ["qwdota", "wdota"]):
-      # For Zvdota FP8 instructions, we need to encode the FP8 type and output
-      # type to disambiguate between f8e4m3 and f8e5m2 which both use vuint8
+    elif any(op in name
+             for op in ["qwdota", "wdota", "qwbdota", "wbdota", "fbdota"]):
+      # For Zvdota/Zvbdota FP8 instructions, we need to encode the FP8 type
+      # and output type to disambiguate between f8e4m3 and f8e5m2 which both
+      # use vuint8
```
hdelassus: Can we try using `f8e5m2` instead of `alt`?

I remember [this discussion was had](https://github.com/riscv-non-isa/riscv-rvv-intrinsic-doc/pull/399#issuecomment-3055374514) for BF16 intrinsics that were otherwise ambiguous

topperc: This alt is part of the instruction name not the altfmt field. Can you make a naming suggestion in auto-generated/policy_funcs/overloaded_intrinsic_funcs/20_zvfqwbdota8f_-_fp8_batched_quad-widening_dot_product.adoc

hdelassus: Yes sorry my comment wasn't clear.

hdelassus: Ok now you got me thinking, since it is part of the mnemonic name, then it's probably best as-is.

auto-generated/policy_funcs/overloaded_intrinsic_funcs/20_zvfqwbdota8f_-_fp8_batched_quad-widening_dot_product.adoc
```diff
@@ -0,0 +1,224 @@
...
+                                                 size_t ci, size_t vl);
+vfloat32m4_t __riscv_vfqwbdota_alt_f8e4m3_f32_tu(vfloat32m4_t vd,
+                                                 vuint8m8_t vs2, vuint8m1_t vs1,
+                                                 size_t ci, size_t vl);
+vfloat32m8_t __riscv_vfqwbdota_alt_f8e4m3_f32_tu(vfloat32m8_t vd,
+                                                 vuint8m8_t vs2, vuint8m1_t vs1,
+                                                 size_t ci, size_t vl);
+vfloat32m1_t __riscv_vfqwbdota_f8e5m2_f32_tu(vfloat32m1_t vd, vuint8m8_t vs2,
+                                             vuint8m1_t vs1, size_t ci,
+                                             size_t vl);
```
hdelassus: Do we want to have `alt` for `vs1` or be explicit and use `f8e5m2` at the expense of not matching 1:1 the mnemonics?
FYI @ebavier

topperc: If we put the type for vs1 instead of alt, should we also put the type for vs1 for vqwbdotas and vqwbdotau which have the signedness of vs1 in the mnemonic and the signedness of vs2 in altfmt.

ebavier: I think we should avoid using `_alt` anywhere in intrinsics names in favor of a more descriptive type tag, including in signed and unsigned intrinsics.  Whether the "alt" occurs in `vtype` or instruction mnemonic is irrelevant; both are implementation details not suitable for exporting to intrinsics users.

I think we should also more seriously consider adding actual vector types for the 8-bit floating-point formats.

aswaterman: I agree with @ebavier.  The use of `altfmt` is an encoding hack, much like placing `vsew` in `vtype`.  (If instruction encoding space and code size didn't matter, we wouldn't have `vtype` at all.)  The intrinsic name should reflect the operation, not the encoding scheme.

topperc: The current proposal uses types in the intrinsic name for `altfmt`. It's only the `alt` and `s`/`u` in the mnemonic names that ended up in the intrinsic names. We've never strayed from mnemonic names before.

We have no precedent for 3 types in the intrinsic name so far, but I guess that's what we're suggesting here?

aswaterman: I was only talking about e.g. `__riscv_vfqwbdota_alt_f8e4m3_f32_tu` above, where I'd use `f8e5m2` in place of `alt`.

`alt` is indirectly naming a type here, so one could argue that we're breaking with precedent either way.

4vtomat: One remaining issue is that should we encode as `{vs2_type}_{vs1_type}_{vd_type}` or `{vs1_type}_{vs2_type}_{vd_type}`?

topperc: > One remaining issue is that should we encode as `{vs2_type}_{vs1_type}_{vd_type}` or `{vs1_type}_{vs2_type}_{vd_type}`?

Probably `{vs2_type}_{vs1_type}_{vd_type}` so at least the source operands are in order. Is there any point to have overloaded intrinsics for these? Should we be having this naming conversation on the non-overloaded versions?

ebavier: I agree with `{vs2}_{vs1}_{vd}` ordering for type tags; we already have the `<in>_<out>` convention, and the two in tags should probably match the ordering of the arguments.

Overloads still make sense here to abstract LMUL values.  They would be more powerful with real vector types that could be matched on.  Could we decide to just not do overloaded functions until then?

4vtomat: Yes, we can just remove overloaded intrinsics for now and we'll discuss about the better solution for adding clang type for fp8, thanks!

4vtomat:
Hey @aswaterman are we going to support fp16 version of `Zvfwbdota16bf`? I'm thinking if we want to keep the suffix `{input_type}_{output_type}` instead of just `{output_type}` to make it extendable for fp16 types.

aswaterman:
Although we haven't proposed FP16 variants, they remain a possibility, so I agree there might be some virtue in including both datatypes in the name.

auto-generated/overloaded_intrinsic_funcs/22_zvfbdota32f_-_fp32_batched_dot_product.adoc
```diff
@@ -0,0 +1,6 @@
+
+=== Zvfbdota32f - FP32 Batched Dot Product
+
+[[overloaded-zvfbdota32f-fp32-batched-dot-product]]
+==== Zvfbdota32f - FP32 Batched Dot Product
+Intrinsics here don't have an overloaded variant.
```
topperc: Why don't we have an overload for this? It isn't ambgiguous. I thought we were only dropping overloads for fp8. @ebavier @hdelassus

4vtomat: I was thinking to support them at once when we define c type for ofp8, but yes we can support partial first😅

4vtomat: Do you think maybe we can also remove non-overloaded version until we support ofp8 c type?
otherwise we need to change the interface again

topperc: What is the timeline for having ofp8 c type? What is the plan for existing intrinsics that use ofp8 types like Zvofp8min when we add the ofp8 types?

4vtomat: I have a PR to support ofp8 RVV type without scalar type: https://github.com/llvm/llvm-project/pull/191349
I'm going to change the spec to update zvfofp8min interfaces to use new ofp8 RVV types.

4vtomat:
Remove stacking on https://github.com/riscv-non-isa/riscv-rvv-intrinsic-doc/pull/427 so that it's easier to review

4vtomat:
gentle ping~

4vtomat:
ping again~

4vtomat:
ping~

topperc (approved):
LGTM

PR #429, opened 2026-03-19
Antora refactor

wmat:
Adding Antora requirements so that the spec can be added to docs.riscv.org


doc/rvv-intrinsic-spec.adoc
```diff
@@ -1,9 +1,11 @@
+[[intro]]
 == Introduction
 
 The RISC-V vector C intrinsics provide users interfaces in the C language level to directly leverage the RISC-V "V" extension cite:[riscv-v-spec] (also abbreviated as "RVV"), with assistance from the compiler in handling instruction scheduling and register allocation. The intrinsics also aim to free users from responsibility of maintaining the correct configuration settings for the vector instruction executions.
 
 This document uses the term "RVV" as an abbreviation for the RISC-V "V" extension. This document uses the term "the RVV specification" to indicate the RISC-V "V" extension specification.
 
+[[tesst-macro]]
```
dzaima, line 8: typo

modules/ROOT/attachments/v-intrinsic-spec.pdf
dzaima, line 1: This is a 7.8MB file that every clone of the repo will have to download.

And in the commit history is a 20MB version of this file that'll be downloaded too (unless one clones with `--depth=1` or similar; or if the PR is squash-merged I guess) but which will probably never be looked at by anyone. At the absolute very least it'd be reasonable to squash that away imo.

And of course if this file ever needs to be updated, that'll add more size every time it is. (and no chance for git's delta compression to do much as the PDFs are already compressed)

In total, a clone of this PR branch results in a 125MB `.git` folder, compared to 103MB of `main`, i.e. a total repo size increase of 21%. (I am but a driveby commenter, but this feels worthy of at least a note)

wmat: Yes, you're correct.  Perhaps this requires further thought.

PR #435, opened 2026-07-28
Fix vwabda[u] operand order

garthlei:
The current vwabda[u] intrinsics are using the vmacc-like `vs1, vs2` operand order, while the asm operand order in the current Zvabd specification is `vs2, vs1`. I guess we should be consistent with the Zvabd spec?


topperc (approved):
LGTM

garthlei:
@topperc @4vtomat @kito-cheng Could anyone merge this?

